PR #59, opened 2021-11-01
feat: add extension not found utility functions

genslein:
Adds 3 utility functions for downstream clients to validate when using `ExtensionString` to check compiled vervet specification documents. 

* One for the stability extension
* One for the version extension
* And a generic so that the logic doesn't have to be repeated.

### DIFF
--- a/resource.go
+++ b/resource.go
@@ -3,6 +3,7 @@ package vervet
 import (
 	"context"
 	"encoding/json"
+	"errors"
 	"fmt"
 	"path/filepath"
 	"sort"
@@ -31,6 +32,19 @@ type Resource struct {
 	Name         string
 	Version      Version
 	sourcePrefix string
+}
+
+type extensionNotFoundError struct {
+	extension string
+}
+
+func (e *extensionNotFoundError) Error() string {
+	return fmt.Sprintf("extension \"%s\" not found", e.extension)
+}
+
+func (e *extensionNotFoundError) Is(err error) bool {
+	_, ok := err.(*extensionNotFoundError)
+	return ok
 }
 
 // Validate returns whether the Resource is valid. The OpenAPI specification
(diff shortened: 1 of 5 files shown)


resource.go
```diff
@@ -166,6 +166,24 @@ func ExtensionString(extProps openapi3.ExtensionProps, key string) (string, erro
 	}
 }
 
+// IsExtensionNotFound returns bool whether error from ExtensionString is not found versus unexpected.
+func IsExtensionNotFound(err error, extension string) bool {
+	if err != nil && err.Error() == fmt.Sprintf("extension \"%s\" not found", extension) {
```
cmars, line 171: I think the type-system can be leveraged here a bit more with a custom error type.

Instead of returning fmt.Errorf errors above in ExtensionString, we could return a new type (`extensionNotFoundError`?) that implements `Error()` to satisfy the `error` interface and format the message. You could then type-check for this error directly (`extErr, ok := err.(*extensionNotFoundError)`) to check if the error matches.

If we find that we need more error classification, we could adopt a scheme like https://github.com/jbowes/vice -- but a simple type check like might be fine to start, especially if the error concrete types are kept private.

As another example, how the stdlib implements os.IsNotExist: https://cs.opensource.google/go/go/+/refs/tags/go1.17.2:src/os/error.go;l=92

genslein: Makes sense, I'll refactor for the proper modeling here. Update shortly.

resource.go
```diff
@@ -160,12 +177,17 @@ func ExtensionString(extProps openapi3.ExtensionProps, key string) (string, erro
...
+			return "", &extensionNotFoundError{key}
 		}
 		return "", fmt.Errorf("unexpected extension %v type %T", m, m)
 	}
 }
 
+// IsExtensionNotFound returns bool whether error from ExtensionString is not found versus unexpected.
+func IsExtensionNotFound(err error) bool {
+	return errors.Is(err, &extensionNotFoundError{})
+}
```
cmars, line 189: Please add test cases:
- err is something else `fmt.Errorf("something completely different")`
- err is nil
- err is a wrapped extensionNotFoundError

cmars (commented):
Looking good, minor test request below.

cmars (approved):
Thanks for this!

PR #61, opened 2021-11-08
Adds prompt to overwrite when rerunning scaffold init

### LINKED ISSUE #56: `vervet scaffold init` should display a more helpful message when a project is already initialized
If a project is already initialized, it should display a helpful message like:

your new project is ready (here's what to do next!)
use -f|--force to overwrite existing files

and exit 0 (be otherwise idempotent). And/or... an interactive prompt (Your project is already init'ed, overwrite these files (y/N):...)

Right now we get:

```
$ vervet scaffold init path/to/scaffold

2021/10/29 11:08:08 open .vervet.yaml: file exists
```

jaycee:
Currently running `scaffold init` a second time results in an error
message being propogated to the user as the cli output. This is
harmless, as the error regards the files being created already existing,
but not a great UX.

This adds promptui to run an interactive prompt for overwriting the
files. If the user selects to overwite, the command behaves as though it
were run with the `-force` option. If the user doesn't, it ends the
command but doesn't return an error.

Other related changes introduced by this change:

 * The cli app is wrapped in a new VervetApp struct, allowing config
   options specific to vervet but not to the CLI interface to be set--
   one of these options is a prompting interface.
 * The vervet app is added to the context passed into the cli app, so
   that vervet specific info can be used in the commands called by the
   cli app.

Fixes #56


cmd/scaffold.go
```diff
@@ -34,11 +37,40 @@ func ScaffoldInit(ctx *cli.Context) error {
...
 	err = sc.Init()
 	if err != nil {
 		return err
 	}
 	return nil
 }
+
+func promptForOverwrite() bool {
+	prompt := promptui.Prompt{
+		Label:   "Scaffold already initialed; do you want to overwrite (y/N)?",
```
cmars, line 62:
```suggestion
		Label:   "Scaffold already initialized; do you want to overwrite (y/N)?",
```

cmd/scaffold.go
```diff
@@ -34,11 +37,40 @@ func ScaffoldInit(ctx *cli.Context) error {
...
+		Default: "N",
+		Validate: func(input string) error {
+			input = strings.ToLower(input)
+			if input != "y" && input != "n" {
+				return errors.New("invalid response: must be 'y' or 'n'")
+			}
+			return nil
+		},
+	} // Not using IsConfirm; it seems to have rendering issues in a few terminal cases.
+	if result, err := prompt.Run(); err == nil && result == "y" {
```
cmars, line 72: Should we propagate the error if `err != nil` ?

cmars (approved):
LGTM with a couple of minor things.

cmd/scaffold.go
jaycee: We can; I had thought that in the event where Prompt.Run failed something weird enough happened that we should just do nothing (e.g. return `nil`) but that does obfuscate that something has indeed gone wrong.

I'll update to return the error; when expanding the use of promptui maybe I can put together something better to handle `Run` errors than just returning it to the the user.

cmd/scaffold.go
```diff
@@ -34,7 +35,27 @@ func ScaffoldInit(ctx *cli.Context) error {
...
 	err = sc.Organize()
 	if err != nil {
-		return err
+		// If we encounter an error setting up the project files, see if the user
+		// wants to overwrite them; if not, return w/o error to end running the
+		// scaffold init command.
+		vervetApp, ok := ctx.Context.Value(vervetKey).(*VervetApp)
+		if !ok {
+			return errors.New("could not retrieve vervet app from context")
+		}
```
cmars, line 44: A `func AppFromContext(ctx context.Context) *VervetApp` helper might be nice for this. If not in this PR, please leave a TODO. `vervet version new` also needs a prompting capability, it can be refactored when we get to that one.

cmd/scaffold.go
```diff
@@ -34,7 +35,27 @@ func ScaffoldInit(ctx *cli.Context) error {
 	}
 	err = sc.Organize()
 	if err != nil {
```
cmars, line 37: I'm wondering if `os.IsExist` would be a more precise test for whether we need to prompt. It'd be nice if we only prompt on a failure to overwrite, and not other things (like no file permissions) to avoid confusing behavior.

cmars (approved):
LGTM with a couple thoughts.

cmd/scaffold.go
jaycee: I'm strongly in agreement with this idea, thanks! I've left the TODO for the follow up. A `contextWithApp` helper might be cleaner too where it's set up in `cmd/cmd.go`

cmd/scaffold.go
jaycee: `os.IsExist` can't actually be used here; the error being returned doesn't unwrap to one that matches for that.

Instead I've added a custom error--`ErrAlreadyExists`--in `internal/scaffold/scaffold.go`, and added logic to check if the files already exist as we're copying them over and return that, which is then used as the key for running the prompt.

cmars (approved):
Ship it!

PR #67, opened 2021-11-10
fix: prevent empty version folders from being created

gablaxian:
The current method to creating the version spec files involved looping over all identified versions and then over the stabilities and creating folders for all, regardless of whether a spec file was necessary.

While not a problem, per se, it conflicted with the mapping of those versions within Registry when looping over the compiled versions and checking for their existence. In that respect, if a folder is found with no file, then that should be a legitimate error.

The solution, simply, is to only create folders for which spec files exist.


internal/compiler/compiler_test.go
```diff
@@ -63,7 +63,7 @@ func TestCompilerSmoke(t *testing.T) {
 	c := qt.New(t)
 	setup(c)
 	ctx := context.Background()
-	outputPath := c.Mkdir()
+	outputPath := testing.TB.TempDir(t)
```
gablaxian, line 66: My IDE told me that the other method was deprecated and this was the preferred approach 🤷

cmars: It is. A shorter way is `c.TempDir(t)`; `quicktest.C` wraps `testing.TB`.

gablaxian: It appears I hadn't pulled latest changes. Kept the original.

internal/compiler/compiler.go
```diff
@@ -286,18 +286,24 @@ func (c *Compiler) Build(ctx context.Context, apiName string) error {
...
 				if err == vervet.ErrNoMatchingVersion {
 					continue
 				} else if err != nil {
 					return buildErr(err)
 				}
 
+				// Create the directories, but only if a spec file exists for it.
+				versionDir := api.output.path + "/" + version.String()
+
+				if spec != nil {
```
cmars, line 300: I don't think `spec` can be `nil` here because we return early on `err != nil` above. Generally our methods will return a valid non-nil result when a possible error return value is nil. This helps keep the level of indent low.

internal/compiler/compiler_test.go
```diff
@@ -102,6 +102,15 @@ func TestCompilerSmoke(t *testing.T) {
...
 	c.Assert(err, qt.IsNil)
 
+	// Verify created files/folders are as expected
+	// Look for existence of /2021-06-01~experimental
+	_, err = os.Stat(outputPath + "/2021-06-01~experimental")
+	c.Assert(err, qt.IsNil)
+
+	// Look for absence of /2021-06-01
+	_, err = os.Stat(outputPath + "/2021-06-01")
+	c.Assert(err, qt.Not(qt.IsNil))
```
cmars, line 112: A little more precise:

```suggestion
	c.Assert(os.IsNotExist(err), qt.IsTrue)
```

cmars (approved):
Thanks for this! LGTM, just a couple of small suggestions.

internal/compiler/compiler.go
gablaxian: I only used `nil` because I output this and saw it returned `nil` when no spec existed. I can dig a bit deeper?

cmars: Ah, looking into this, I realize now you are indeed correct, this _can_ be nil, looking at `loadResource` in `resource.go`. If there are no paths in the OpenAPI doc, a nil spec can propagate up.

So here I think it's fine as-is!  And there's possibly a bug to followup on, or at least godoc-comment better.

testdata/output/2021-06-04~experimental/spec.json
```diff
@@ -457,7 +457,7 @@
         }
       },
       "x-snyk-api-resource": "hello-world",
-      "x-snyk-api-version": "2021-06-01"
+      "x-snyk-api-version": "2021-06-01~experimental"
```
genslein, line 460: This is somewhat puzzling to me as I'd expect this to be `2021-06-04~experimental`

genslein:
cc @Todai88 I believe this would address https://github.com/snyk/v3-go-libs/blob/main/oas_router/oas_router.go#L244-L249 correct my if I'm mistaken @gablaxian @cmars ?

testdata/output/2021-06-04~experimental/spec.json
cmars: `/examples/hello-world/{id}` in this version comes from `2021-06-01~experimental`. This version exists because of the projects addition in `2021-06-04~experimental`.

cmars: Below you'll find that the projects path has extensions reflecting its introduction:

```
      "x-snyk-api-resource": "projects",
      "x-snyk-api-version": "2021-06-04~experimental"
```

So with this change, there will be at least one path at the version of the compiled spec. Other paths in the file may be older and/or of greater stability -- it's a snapshot of "what versions are effective at this version".

gablaxian:
> cc @Todai88 I believe this would address https://github.com/snyk/v3-go-libs/blob/main/oas_router/oas_router.go#L244-L249 correct my if I'm mistaken @gablaxian @cmars ?

I'm not totally sure. What this PR is fixing, is this error which you see on registry startup:
```
{"name":"Error","message":"ENOENT: no such file or directory, open '/home/circleci/registry/dist/web/routes/api/sarif/versions/2021-08-13~beta/spec.json'"
```
which occurs because folders for all stabilities are created per version, regardless of whether the endpoint exists at that stability level. So when you first start creating endoints, you are going to end up with a lot of redundant folders. And when we scan to get all the spec files, we error because we expect a file in every folder.

genslein:
> I think the issue you linked to is something else and, if I am not mistaken, is intended?
Yes, it was a different issue being confusing generating many folders and files more than the pre-compiled specs define. ignore me 🏃

PR #69, opened 2021-11-17
feat: adds interactive prompting to version new cmd

### LINKED ISSUE #57: Improve `vervet version new` UX
Creating a new resource in a newly scaffolded API is somewhat difficult, because you don't immediately know what the new API is named, which you need in order to say:

`vervet version new my-api my-new-resource`

Ideas:

- Make the API argument optional -- it will likely have to be specified with a flag instead, or prompted for
  (Prompts could help improve #56 as well -- what's a good package for this?)
- `vervet version list` should list empty APIs -- it currently doesn't

An interactive CLI with flags to make it non-interactive seems the best way forward here...

jcsackett:
The version command currently requires API and resource to be specified at
invocation. It would be better if the command interactively prompted for those
values if they were omitted by the user.

Fixes #57

### DIFF
--- a/cmd/cmd.go
+++ b/cmd/cmd.go
@@ -31,17 +31,31 @@ type VervetApp struct {
 
 // VervetPrompt defines the interface for interactive prompts in vervet.
 type VervetPrompt interface {
-	Confirm(label string) (bool, error)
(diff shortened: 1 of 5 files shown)


cmars (approved):
LGTM!

PR #73, opened 2021-11-22
feat: add Optic CI linter

cmars:
Add support for linting OpenAPI spec file changes with Optic CI. This
linter operates on a comparison of two versions of the spec file, a
--from and --to version.

Files may be sourced from Git or a local working copy.


internal/optic/linter.go
```diff
@@ -0,0 +1,226 @@
...
+		image:      image,
+		fromSource: fromSource,
+		toSource:   toSource,
+		runner:     &execCommandRunner{},
+	}, nil
+}
+
+// Run runs Optic CI on the given paths. Linting output is written to standard
+// output by Optic CI. Returns an error when lint fails configured rules.
+func (l *Optic) Run(ctx context.Context, paths ...string) error {
```
jcsackett, line 187: `l` for 'lint' instead of `o` for 'Optic'? Is `o` commonly expected to be another type, in the way that `i` is commonly an increasing integer?

internal/optic/linter.go
```diff
@@ -0,0 +1,226 @@
...
+		args = append(args, "--from", fromFile)
+	}
+	toFile, err := l.toSource.Fetch(path)
+	if err != nil {
+		return err
+	}
+	if toFile != "" {
+		args = append(args, "--to", toFile)
+	}
+	// TODO: provide rule context
```
jcsackett, line 217: Can we provide info on what the options in cmd do? Or a link from here to the docs for the linter being run, so it's easier to figure out exactly what's being run when revisiting this code in the future? Or is that what is meant by the TODO?

internal/optic/linter_test.go
```diff
@@ -0,0 +1,162 @@
...
+
+	qt "github.com/frankban/quicktest"
+	"github.com/go-git/go-git/v5"
+	"github.com/go-git/go-git/v5/plumbing"
+	"github.com/go-git/go-git/v5/plumbing/object"
+	"github.com/google/uuid"
+
+	"github.com/snyk/vervet/testdata"
+)
+
```
jcsackett, line 20: Noticing throughout these tests you're using `defer` for what looks like cleanup, rather than `t.Cleanup`; I think for anything that is test cleanup we probably should use that function, even if defer works, so we don't get into a weird situation if we start moving things around or creating more complicated tests/test helpers. What do you think?

internal/optic/linter_test.go
```diff
@@ -0,0 +1,162 @@
...
+	err = l.Run(ctx, "spec.yaml")
+	c.Assert(err, qt.IsNil)
+	c.Assert(runner.runs[0], qt.Contains, "--from")
+	c.Assert(runner.runs[0], qt.Contains, "--to")
+
+	// Command failed.
+	runner = &mockRunner{err: fmt.Errorf("nope")}
+	l.runner = runner
+	err = l.Run(ctx, "spec.yaml")
+	c.Assert(err, qt.ErrorMatches, "nope")
```
jcsackett, line 123: A missed opportunity for `bad wolf` to rear its head again 😉

jcsackett (approved):
LGTM, with a few questions/comments. 👍🏼

internal/optic/linter.go
cmars: More like `l` for lazy :P. I think I should rename this.

internal/optic/linter.go
cmars: Improved the existing TODO a bit, and added another TODO for what you're asking for. It's a command that is still being developed elsewhere in Sweater Comb's `optic-main` branch -- and its dependencies.

PR #74, opened 2021-11-22
feat: add Optic CI linter to config

cmars:
This change adds support for configuring the Optic CI linter.

Drive-by: Break up config.go into separate files.


jcsackett (approved):
LGTM!

PR #76, opened 2021-11-24
feat: introduce vervet-underground

Todai88:
Readme should tell us what we want.


jcsackett (commented):
LGTM! 

Using config.json for configuration is a little odd in a go app, isn't it? Is that a company standard rather than the more common yaml or toml config paths most go projects seem to pursue?

jcsackett (approved):

Todai88:
> LGTM!
> 
> Using config.json for configuration is a little odd in a go app, isn't it? Is that a company standard rather than the more common yaml or toml config paths most go projects seem to pursue?

I'm not sure if it's a company standard, but the `config` library that we created in V3 suggests it maybe is. 
Arguably JSON is a bit easier to parse than YAML, but I'm sure we could extend the lib to use YAML too, if we wanted to :)

PR #80, opened 2021-11-30
wip: start polling

Todai88:


cmars (changes requested):
I'd advise against scraping synchronously in an HTTP request, it may be a long running operation. A periodic background scrape per hour using `internal/scraper`, and running immediately on startup, is probably good enough to begin with.

PR #81, opened 2021-11-30
fix: provide optic-ci with context

cmars:
In order for optic-ci to evaluate versioning lifecycle rules, some
external information needs to be provided, telling optic:

* When a proposed change would occur
* The resource and version being changed
* Existing versions in the API that may deprecate the version under
  change.


jcsackett (commented):
LGTM!

jcsackett (approved):

PR #82, opened 2021-11-30
fix: restrict optic configuration to supported functionality

cmars:
Optic rules are not currently applied to the compiled output stage and
this is not yet supported, lacking a clear use case or definition.

Optic rules are also not allowed for "overrides". Linter overrides
were a stop-gap measure that should be better solved with a more
well-defined standards & compliance lifecycle.

Add some additional config sanity checks.


jcsackett (approved):
LGTM!

PR #83, opened 2021-11-30
docs: improve intro with why and what VU is about

cmars:
CC @Todai88 @genslein


Todai88 (approved):
LGTM. 

Good to have something that we can refer back to! :)

PR #88, opened 2021-12-04
feat: implement VU service scraper

cmars:
Add a scraper which fetches all the OpenAPI versions from each service
and updates them in VU storage.

Add an in-memory storage implementation to serve as a functional
reference.

Drive-by: consolidated config package, added godocs


genslein (commented):
https://docs.aws.amazon.com/AmazonS3/latest/userguide/object-tagging.html
https://docs.aws.amazon.com/AmazonS3/latest/userguide/tagging-managing.html

We can definitely use custom tags for our object uploads.

vervet-underground/internal/storage/etag.go
```diff
@@ -0,0 +1,14 @@
...
+package storage
+
+import (
+	"crypto/sha256"
+	"encoding/base64"
+)
+
+// ETag returns the content hash that VU storage and scraped services agree on for OpenAPI content.
+func ETag(contents []byte) string {
+	buf := sha256.Sum256(contents)
```
genslein, line 12: https://docs.aws.amazon.com/AmazonS3/latest/API/RESTCommonResponseHeaders.html ETags are mix of things, not a straight Sha256 

https://teppen.io/2018/06/23/aws_s3_etags/#what-is-an-s3-etag
https://stackoverflow.com/questions/12186993/what-is-the-algorithm-to-compute-the-amazon-s3-etag-for-a-file-larger-than-5gb

genslein: If we decide to use KMS encryption, For the ETag, you'd need to calc MD5 with S3 library prior to upload https://stackoverflow.com/a/53886736

cmars: The ETag I'm considering here would be something agreed between VU and the services it scrapes -- the idea being that we can `HEAD /openapi/{version}` and then quickly decide if we need to scrape and store a new revision of that release version.

ETags between VU and the storage backend would be different.

genslein: Can we call it something else then? This seems really confusing to differentiate types of ETags

genslein: Maybe `ShaTag` or `OASTag`?

genslein (changes requested):
Rename the VU <-> Service SHA not ETag to not confuse it with Backing storage ETag

cmars:
Use of ETag is prescribed for HEAD requests though, https://httpwg.org/specs/rfc7234.html#head.effects.

I think it should be fine to use one ETag for VU <-> service requests, and another ETag for VU <-> AWS S3 requests. They are using the same response header but in entirely different contexts.

cmars:
Looks like `Digest:` is a much better fit for this. I'll update!

https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Digest
https://datatracker.ietf.org/doc/html/draft-ietf-httpbis-digest-headers-05#section-3

vervet-underground/internal/scraper/scraper.go
```diff
@@ -0,0 +1,245 @@
...
+	for i := range s.services {
+		svc := s.services[i]
+		wg.Add(1)
+		go func() {
+			defer wg.Done()
+			errCh <- s.scrape(ctx, scrapeTime, svc)
+		}()
+	}
+	go func() {
+		wg.Wait()
```
genslein, line 115: Not sure i understand why the WG wait here is in another goroutine. I'd expect the `Run` function here to exit without actually waiting for all the scrapers to finish by this

vervet-underground/internal/scraper/scraper.go
```diff
@@ -0,0 +1,245 @@
...
+		// Not supporting HEAD is fine, we'll just come back with a GET
+		return true, nil
+	}
+	if resp.StatusCode != http.StatusOK {
+		return false, httpError(resp)
+	}
+	if ct := resp.Header.Get("Content-Type"); ct != "application/json" {
+		return false, errors.Errorf("unexpected content type: %s", ct)
+	}
+	digest := storage.DigestHeader(resp.Header.Get("Digest"))
```
genslein, line 239: ❤️‍🔥 🤠

vervet-underground/internal/scraper/scraper_test.go
```diff
@@ -0,0 +1,153 @@
...
+		contents: map[string]string{
+			"2021-10-01": `{"paths":{"/geckos": {}}}`,
+			"2021-10-16": `{"paths":{"/geckos": {}, "/puppies": {}}}`,
+		},
+	}
+	animalsService := httptest.NewServer(animals.Handler())
+	c.Cleanup(animalsService.Close)
+
+	tests := []struct {
+		service, version, etag string
```
genslein, line 69:
```suggestion
		service, version, digest string
```

vervet-underground/internal/scraper/scraper_test.go
```diff
@@ -0,0 +1,153 @@
...
+	sc, err := scraper.New(cfg, st, scraper.Clock(func() time.Time { return t0 }))
+	c.Assert(err, qt.IsNil)
+
+	// Cancel the scrape context after a timeout so we don't hang the test
+	ctx, cancel := context.WithTimeout(context.Background(), time.Second*10)
+	c.Cleanup(cancel)
+
+	// No version digests should be known
+	for _, test := range tests {
+		ok, err := st.HasVersion(ctx, test.service, test.version, test.etag)
```
genslein, line 91:
```suggestion
		ok, err := st.HasVersion(ctx, test.service, test.version, test.digest)
```

vervet-underground/internal/storage/mem/mem.go
```diff
@@ -0,0 +1,86 @@
...
+// NotifyVersions implements scraper.Storage.
+func (s *Storage) NotifyVersions(ctx context.Context, name string, versions []string, scrapeTime time.Time) error {
+	s.mu.Lock()
+	defer s.mu.Unlock()
+	// TODO: implement notify versions; update sunset when versions are removed
+	return nil
+}
+
+// HasVersion implements scraper.Storage.
+func (s *Storage) HasVersion(ctx context.Context, name string, version string, etag string) (bool, error) {
```
genslein, line 56:
```suggestion
func (s *Storage) HasVersion(ctx context.Context, name string, version string, digest string) (bool, error) {
```

genslein (approved):
Some of the variable names could be moved to `digest` but won't block

vervet-underground/internal/scraper/scraper.go
cmars: `Run` is intentionally synchronous. Initial use case for this will be running a background periodic batch job; it's nice to wait for all the scrapes to finish in the main goroutine. For fire-and-forget or async usage, one could `go scraper.Run(...)`.

The `wg.Wait` in the goroutine is used to close the errors channel exactly once, when all the scrapers are sure to be complete (you have to be sure, because sending on a closed channel will panic). This is done in a goroutine so that the main goroutine can aggregate the errors by iterating the errors channel and then return it directly.

Issue #78, opened 2021-11-29
Annotate paths with a version changelog


cmars:
API documentation needs to represent a "changelog" per resource endpoint: a list of versions in which a change to a given resource endpoint was released.

This can be represented as an OpenAPI extension on the PathInfo element. For example, a response to `GET /openapi/experimental` might show:

```
...
paths:
  /some_resource:
...
    x-snyk-api-version: '2021-08-09~experimental'
    x-snyk-api-releases: ['2021-06-07~experimental', '2021-07-21~beta', '2021-08-09~experimental']
```

Indicating that /some_resource has 3 versioned releases. The latest experimental is the one shown, but if one wanted to compare this against prior changes, they could look at `2021-06-07~experimental` or `2021-07-21~beta`.

To implement this (CC @mrded @jcsackett), I would look at adding the extensions to paths in `vervet.LoadResourceVersionsFileset`. After the sort, you'll have everything you need:
* The OpenAPI specs at each resource version relevant to this particular resource.
* Sorted in version order, ascending.

The dead-simplest approach would be to set `x-snyk-api-releases` to a slice containing this list of versions, at each version. That might be slightly redundant in cases where the only release change was an endpoint being added or removed, since this list of versions is _per-resource_ rather than _per-endpoint_. In practice, that may not matter much -- it could be something to refine later, or we might actually _want_ to highlight that the resource changed, while this particular endpoint did not.

cmars:
Fixed in #89

PR #89, opened 2021-12-05
feat: annotate each path with other release versions

cmars:
API documentation needs to represent a "changelog" per resource
endpoint: a list of versions in which a change to a given resource
endpoint was released.

This change adds a list of other resource release versions in which a
particular path was declared for such navigation purposes.

Implements #78.

### DIFF
--- a/vervet-underground/internal/scraper/scraper.go
+++ b/vervet-underground/internal/scraper/scraper.go
@@ -8,7 +8,6 @@ import (
 	"io/ioutil"
 	"net/http"
 	"net/url"
-	"sync"
 	"time"
 
 	"github.com/pkg/errors"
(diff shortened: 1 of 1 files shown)


resource.go
```diff
@@ -139,31 +144,49 @@ func LoadResourceVersions(epPath string) (*ResourceVersions, error) {
...
+	for _, releases := range pathReleases {
+		sort.Sort(releases)
+	}
+	// Sort the resources themselves by version
+	sort.Sort(resourceVersionSlice(resourceVersions.versions))
+	// Annotate each path in each resource version with the other change
+	// versions affecting the path. This supports navigation across versions.
+	for _, rc := range resourceVersions.versions {
+		for path, pathInfo := range rc.Paths {
+			pathInfo.ExtensionProps.Extensions[ExtSnykApiReleases] = pathReleases[path].Strings()
```
jcsackett, line 186: Is there a bit of code that's automatically adding extension sections to the spec output? I see the data added here, and I see a string representation utility function at the end of the PR, but no updates to have this put in spec, so guessing that's automatic?

resource.go
```diff
@@ -139,31 +144,49 @@ func LoadResourceVersions(epPath string) (*ResourceVersions, error) {
 }
 
 func LoadResourceVersionsFileset(specYamls []string) (*ResourceVersions, error) {
-	var eps ResourceVersions
+	var resourceVersions ResourceVersions
```
jcsackett, line 147: Good renaming; what was `eps` for?

jcsackett (commented):
LGTM! Questions in the diff for my own edification (unless it turns out I stumbled across something missing 🙂)

resource.go
cmars: Endpoints.

resource.go
cmars: These extensions are basically being added on load, which gets merged into the compiled result by `vervet.Merge`.

The utility function is used to read the value of extensions provided in the source OpenAPI spec.

cmars (commented):
Replies to comments.

Issue #86, opened 2021-12-02
Add sunset information to compiled OpenAPI


cmars:
When Vervet compiles resource versions, it has a top-level view of all versions of a given resource. When a resource version is deprecated by a newer version, we can annotate the deprecated version with extensions that indicate:

* The newer version that deprecates the deprecated one. (x-snyk-deprecated-by)
* The eligible sunset date of the deprecated one, triggered by the date of the newer version. (x-snyk-sunset-eligible)

Note that the sunset date indicates when the deprecated version _may_ be removed, it does not mean that it _must or necessarily will_ be removed on that date. Post-sunset versions might stick around longer, but they don't have to, and a front-end router may stop routing requests to these.

ianlivingstone:
There are standard headers for sending "this is deprecated, heres more information" that are standardized. We should try and adopt those as much as possible (e.g. we already do this internally, in places).

https://tools.ietf.org/id/draft-dalal-deprecation-header-01.html

cmars:
Depends on #78, #89

cmars:
Fixed in #90

PR #90, opened 2021-12-06
feat: add version deprecation and sunset eligibility

cmars:
Annotate paths in a given resource version with the newer release
version that deprecates it, and when the deprecated version may be
sunset.

This may be used in linting, middleware, and documentation.

Implements #86.

Stacked on #89.

### DIFF
--- a/version.go
+++ b/version.go
@@ -30,6 +30,15 @@ func (v *Version) String() string {
 	return d
 }
 
+// AddDays returns the version corresponding to adding the given number of days
+// to the version date.
+func (v *Version) AddDays(days int) Version {
+	return Version{
+		Date:      v.Date.AddDate(0, 0, days),
+		Stability: v.Stability,
+	}
+}
+
 // Stability defines the stability level of the version.
 type Stability int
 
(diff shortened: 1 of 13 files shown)


version.go
```diff
@@ -146,6 +155,46 @@ func (v *Version) Compare(vr *Version) int {
 	return stabilityCmp
 }
 
+// DeprecatedBy returns true if the given version deprecates the caller target
+// version.
+func (v *Version) DeprecatedBy(vr *Version) bool {
+	return v.Compare(vr) < 0
```
jcsackett, line 161: Why do `< 0` here? The returns from `Compare` are only -1, 0, or 1, so `== -1` makes more sense to me to show it's a particular value.

This is fine by me if your preference is to stick w/ `<` -- I'm curious if we would ever add any other values coming from `Compare`?

jcsackett (approved):
LGTM, thanks! One question in the diff below, more a question of style than any sort of blocker--take it or leave it 🤷🏼 🙂

version.go
cmars: We wouldn't, a direct `== -1` would suffice here. Looking at this more carefully now, I'm now wondering, does a version deprecate itself? I'm thinking it probably shouldn't!

PR #93, opened 2021-12-07
fix: ensures destination components exist in merge

jcsackett:
openapi3 sets the zero value of its schema objects to uninitialized
maps. This is generally fine when loading a spec, as that process
initializes the map, but when we merge items in vervet if a component
had no data in the initial spec the component map remains unitialized,
causing a panic in `Merge` when we attempt to write to it.

To address this, we check at merge time if the src component map is
initialized (not nil) and if the destination map is not. If that's the
case, we call `make` to set the map to a non nil value.

### DIFF
--- a/merge.go
+++ b/merge.go
@@ -44,7 +44,38 @@ func mergeTags(dst, src *openapi3.T, replace bool) {
 	}
 }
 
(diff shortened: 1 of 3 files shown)


cmars (approved):
:shipit: 🚀

PR #94, opened 2021-12-09
feat: service collation algorithm

genslein:
Implements both in-memory storage algorithm
and aggregate service spec merges to the
collated Vervet Underground Spec per
version identified.

Scraper test added for post-processing collation on scraping.


vervet-underground/internal/storage/mem/mem.go
```diff
@@ -6,81 +6,241 @@ package mem
...
+	timestamp time.Time
+	digest    Digest
+	blob      []byte
 	// TODO: store the sunset time when a version is removed
 	//sunset    *time.Time
 }
 
-type serviceVersions map[serviceVersion][]contentRevision
+// MemoryStorage Storage provides an in-memory implementation of Vervet Underground storage.
+type MemoryStorage struct {
```
cmars, line 49: Go names should avoid repetition. Becuase the package is `mem`, `Storage` is unambiguous.

You'll see this pattern repeated throughout the Go standard library. `bufio.Reader`, `crypto/rand.Reader`, `io.Reader`. Consider `http.Header` vs something like `http.HTTPHeader`; you don't need to call it `HTTPHeader` because you always refer to it in its package.

This differs from other languages like Java, Typescript or Rust where you often import symbols directly out of packages into the current context. Pulling symbols into the current context from another package is possible in Go (`import . "somepackage"`) but it's highly irregular to do so and strongly discouraged.

Not only does this improve clarity, it tends to force package names to be meaningful.

vervet-underground/internal/storage/mem/mem.go
```diff
@@ -6,81 +6,241 @@ package mem
...
-	service string
-	version string
-}
+type Digest string
 
-type contentRevision struct {
-	timestamp time.Time
-	digest    string
+// VersionedResourceMap map [service-name] Vervet Version slice array
+type VersionedResourceMap map[string]vervet.VersionSlice
```
cmars, line 25: Do these types need to be exported outside of `mem`? If not, they shouldn't be. In general, it's good to keep the API surface small where possible.

vervet-underground/internal/storage/mem/mem.go
```diff
@@ -6,81 +6,241 @@ package mem
...
-func (s *Storage) NotifyVersion(ctx context.Context, name string, version string, contents []byte, scrapeTime time.Time) error {
+func (s *MemoryStorage) NotifyVersion(ctx context.Context, name string, version string, contents []byte, scrapeTime time.Time) error {
 	s.mu.Lock()
 	defer s.mu.Unlock()
-	k := serviceVersion{service: name, version: version}
+
 	digest := storage.Digest(contents)
-	if digests, ok := s.contents[k]; ok {
-		if _, ok := digests[digest]; ok {
+	typedDigest := Digest(digest)
```
cmars, line 97: Hmm... makes me think the `Digest` type should be pulled up into the `storage` package, and used there in `storage.Digest(contents []byte) Digest`?

vervet-underground/internal/storage/mem/mem.go
```diff
@@ -6,81 +6,241 @@ package mem
...
+		if err != nil {
+			log.Error().Err(err).Msgf("Could not merge ServiceRevision %s", serviceRevision)
+			return err
+		}
+
+		if dst == nil {
+			dst = src
+		} else {
+			// TODO: use proper specs for testing or this fails hard panic
+			vervet.Merge(dst, src, true)
```
cmars, line 243: #93 might fix panics you've seen here? If not, let's dig into why, could be a bug in merge where we've missed something and hit an uninitialized map. (A drive-by fix there would be great! 🙏)

vervet-underground/internal/storage/mem/mem.go
genslein: very reasonably

vervet-underground/internal/storage/mem/mem.go
genslein: I  believed this was me trying to use a nil pointer for `dst` I'll remove this.

vervet-underground/internal/storage/mem/mem.go
```diff
@@ -6,81 +6,244 @@ package mem
...
+		serviceVersion: name + "_" + version,
+		timestamp:      scrapeTime,
+		digest:         typedDigest,
+		blob:           contents,
+	}
+
+	return nil
+}
+
+func (s *MemoryStorage) CollateVersions() error {
```
cmars: Please add a godoc comment that explains how this is intended to be used, rather than a description of its implementation. That might involve some explanation of what it does, but only in the context of what I need to know in order to use it.

vervet-underground/internal/storage/mem/mem.go
genslein: I need to change the name of the function to not conflict with the type name

vervet-underground/internal/storage/mem/mem.go
cmars: I _think_ what this does is prepare the state of the storage such that I might be able to ask it questions after calling it, from a point-in-time view of all the versions of all the service APIs. And those questions are methods we haven't added yet? Does this sound right?

vervet-underground/internal/storage/mem/mem.go
```diff
@@ -6,81 +6,244 @@ package mem
...
+		err := s.collateVersion(version)
+		if err != nil {
+			log.Error().Err(err).Msgf("Could not Collate Version %s", version.String())
+			errs = multierr.Append(errs, err)
+		}
+	}
+	return errs
+}
+
+func (s *MemoryStorage) GetCollatedVersionSpecs() CollatedVersionMappedSpecs {
```
cmars:
```suggestion
func (s *MemoryStorage) Versions() CollatedVersionMappedSpecs {
```

It may be fair to assume that the versions our storage provides are always collated, if that is its job to do that for us. So perhaps the name can be more concise?

vervet-underground/internal/storage/mem/mem.go
genslein: I can re-expose later, fine to lower case for now as I build

vervet-underground/internal/storage/mem/mem.go
```diff
@@ -6,81 +6,244 @@ package mem
...
+type VersionedResourceMap map[string]vervet.VersionSlice
+
+// MappedRevisionSpecs map [Sha digest of contents string] --> spec contents and metadata
+type MappedRevisionSpecs map[Digest]ContentRevision
 
-type contentRevision struct {
-	timestamp time.Time
-	digest    string
+// CollatedVersionMappedSpecs Compiled aggregated spec for all services at that given version
+type CollatedVersionMappedSpecs map[vervet.Version]openapi3.T
```
cmars: Maybe this can be expressed with a more concise name? It seems like it's a "collection of OpenAPI spec versions indexed by version". Maybe `SpecVersions`?

vervet-underground/internal/scraper/scraper_test.go
```diff
@@ -151,3 +151,63 @@ type errorTransport struct{}
...
+	for _, test := range tests {
+		ok, err := st.HasVersion(ctx, test.service, test.version, test.digest)
+		c.Assert(err, qt.IsNil)
+		c.Assert(ok, qt.IsTrue)
+	}
+
+	err = st.CollateVersions()
+	c.Assert(err, qt.IsNil)
+	collated := st.GetCollatedVersionSpecs()
+	c.Assert(len(collated), qt.Equals, 4)
```
cmars, line 212: It'd be good to assert that the collated versions have the expected contents -- did we collate correctly?

Perhaps this can be spot-checked here, but we'll likely want fixture-based tests with real OpenAPI files for this.

cmars (commented):
Round two.

vervet-underground/internal/storage/digest.go
```diff
@@ -8,6 +8,8 @@ import (
 	"strings"
 )
 
+type Digest string
```
cmars, line 11: Please add a godoc comment.

vervet-underground/internal/storage/digest.go
```diff
@@ -33,8 +35,8 @@ func DigestHeader(value string) string {
 	return ""
 }
 
-// Digest returns the digest of the given contents.
-func Digest(contents []byte) string {
+// GenerateDigest returns the digest of the given contents.
+func GenerateDigest(contents []byte) Digest {
```
cmars, line 39: This is getting more into bikeshed territory, but I wonder if this is really a constructor? I can go either way on this.

```suggestion
// NewDigest returns the digest of the given contents.
func NewDigest(contents []byte) Digest {
```

vervet-underground/internal/storage/mem/mem.go
```diff
@@ -6,81 +6,252 @@ package mem
...
+		collatedVersions:       vervet.VersionSlice{},
+		collatedVersionedSpecs: collatedVersionMappedSpecs{},
 	}
 }
 
 // NotifyVersions implements scraper.Storage.
 func (s *Storage) NotifyVersions(ctx context.Context, name string, versions []string, scrapeTime time.Time) error {
-	s.mu.Lock()
-	defer s.mu.Unlock()
-	// TODO: implement notify versions; update sunset when versions are removed
```
cmars, line 51: Note that this is also the only way you know a sunset version was removed...

vervet-underground/internal/storage/mem/mem.go
```diff
@@ -6,81 +6,252 @@ package mem
...
+		err := s.collateVersion(version)
+		if err != nil {
+			log.Error().Err(err).Msgf("Could not Collate Version %s", version.String())
+			errs = multierr.Append(errs, err)
+		}
+	}
+	return errs
+}
+
+func (s *Storage) GetCollatedVersionSpecs() collatedVersionMappedSpecs {
```
cmars, line 165: You'll need to return a completely distinct copy of this map's contents here (deep, not shallow); you're returning a reference to data that is protected by a mutex elsewhere.

The returned type should be exported. I think a shorter name that describes what the type is (a map of specs keyed by version) rather than how it was made (collated and mapped) would be better.

vervet-underground/internal/storage/mem/mem.go
```diff
@@ -6,81 +6,252 @@ package mem
...
+		if _, ok := revisions[digest]; ok {
 			return nil
 		}
 	} else {
-		s.contents[k] = map[string][]byte{}
+		s.serviceVersionMappedRevisionSpecs[name][version] = mappedRevisionSpecs{}
+	}
+
+	revisions, ok := s.serviceVersionMappedRevisionSpecs[name][version]
+	if !ok {
```
cmars, line 121: We just added this key at [line 117](https://github.com/snyk/vervet/pull/94/files#diff-90da75d2382757cdf44cc3ac56158069e0741cba584ef5347c6afd81624dc558R117) above if it did not exist -- so I don't think this is possible.

Am I missing something?

vervet-underground/internal/storage/mem/mem.go
```diff
@@ -6,81 +6,252 @@ package mem
...
+		s.collatedVersions = append(s.collatedVersions, value)
+	}
+
+	// must sort at end
+	sort.Sort(s.collatedVersions)
+}
+
+func (s *Storage) collateVersion(version vervet.Version) error {
+	// number of services maximum needed
+	contentRevisionCollection := make([]ContentRevision, 0)
```
cmars, line 191:
```suggestion
	contentRevisions := make([]ContentRevision, 0)
```

vervet-underground/internal/storage/mem/mem.go
```diff
@@ -6,81 +6,252 @@ package mem
...
+		}
+	}
+	return errs
+}
+
+func (s *Storage) GetCollatedVersionSpecs() collatedVersionMappedSpecs {
+	return s.collatedVersionedSpecs
+}
+
+func (s *Storage) gatherVersionsRequiredToCollate() {
```
cmars, line 169:
```suggestion
func (s *Storage) updateCollatedVersions() {
```

vervet-underground/internal/storage/mem/mem.go
```diff
@@ -6,81 +6,252 @@ package mem
...
+		} else {
+			// TODO: evaluate whether to use replace bool or not during merging
+			vervet.Merge(dst, src, true)
+		}
+	}
+
+	// lock memory safely here
+	s.mu.Lock()
+	defer s.mu.Unlock()
+	s.collatedVersionedSpecs[version] = *dst
```
cmars, line 255: If you're doing something trivial with a mutex held, and there's no risk of early return or panic, you can:

```suggestion
	s.mu.Lock()
	s.collatedVersionedSpecs[version] = *dst
	s.mu.Unlock()
```

cmars (changes requested):
Implementation-wise this looks very good. Only major concern that needs fixing is the deep-copy when returning the collated specs-by-version.

You might have noticed I'm particular about naming, it's an obsession. Generally I prefer the most concise name that still clearly conveys intent and meaning. Nouns rather than verbs for getters, plural form instead of "-Collection", "-List", etc.

vervet-underground/internal/storage/mem/mem.go
genslein: The naming with reuse of `SpecVersions`, `Version`, and `Doc` makes me concerned about shortening the name here. It's not well understood these are post-Merge ContentRevision Service Specs.

I'm also intentionally strong-typing here to make it obvious not to just put anything in for the keys themselves. 

Happy to talk through a better naming strategy but for the layperson, I'm pretty sure shortening the name will make this a lot less obvious what's happening

vervet-underground/internal/storage/mem/mem.go
genslein: Sunset isn't implemented anywhere. I'd like to separate that out of scope

cmars: I agree it is out of scope, but it is absolutely a TODO to follow up on, please keep the comment. A sunset in a service must propagate to the aggregated SaaS view.

vervet-underground/internal/storage/mem/mem.go
cmars: We can always sort out names later, it is an internal API after all. No worries, just suggestions.

vervet-underground/internal/storage/mem/mem.go
cmars: To be clear, with this issue addressed, I think it's landable. This is the only thing really blocking IMO.

vervet-underground/internal/scraper/scraper_test.go
genslein: Yeah, I think this will have to come from fixtures to validate this from real specs. This was a stopgap to validate the algorithm. I think next is doing the fixtures and integrating with the static service list to target in a subsequent PR

vervet-underground/internal/storage/mem/mem.go
genslein: `Versions` would returning the `collatedVersions vervet.VersionSlice` value. These are the actual OAS.

Like `/openapi` vs. `/openapi/{version}`. It's the latter,  but for each one. I used it as testing function to validate my assumptions. It can be removed and is probably not permanent.

genslein: Again, where naming too short is failing here.

vervet-underground/internal/storage/mem/mem.go
genslein: Yes-ish? This is basically the background processing of all service specs once they are in our memory storage.

This scraping operates in phases.

* Get all versions for all services. Reconcile digest differences for each service's versions if there are non-breaking updates.
* (Where `CollateVersions` comes in) Review all memory records of services to determine a list of unique vervet.Versions. 
  * Use that list to generate VU's collated spec for each version identified. 
  * For each version, search our memory storage to determine if each service has a fuzzy match on the VU version being compiled, and if so, add the latest digest fo compilation
  * Compile each version's ContentRevisions gathered via the search and save to memory

genslein: 🚨 🚨  Bear in mind since this is memory, and not persistent, restarts (deployments/syncs) and containers crashes lose digests.

vervet-underground/internal/storage/mem/mem.go
genslein: removed, I think a leftover from testing

vervet-underground/internal/storage/mem/mem.go
genslein: Added the TODO back in 👍

genslein:
There seems to be a race/flakiness to the Collate test I added, I'm not really sure where it's coming from but I see  the following `github.com/snyk/vervet.mergePaths` was what was giving me trouble before, like there's a race condition setting docs up:

```bash
{"level":"error","error":"no matching version","time":"2021-12-10T17:40:32-05:00","message":"Could not resolve for service http://127.0.0.1:50101 version 2021-09-01"}
{"level":"error","error":"no matching version","time":"2021-12-10T17:40:32-05:00","message":"Could not resolve for service http://127.0.0.1:50101 version 2021-09-16"}
--- FAIL: TestScraperCollation (0.00s)
panic: assignment to entry in nil map [recovered]
        panic: assignment to entry in nil map

goroutine 16 [running]:
testing.tRunner.func1.2({0x1434a80, 0x152c190})
        /usr/local/opt/go/libexec/src/testing/testing.go:1209 +0x24e
testing.tRunner.func1()
        /usr/local/opt/go/libexec/src/testing/testing.go:1212 +0x218
panic({0x1434a80, 0x152c190})
        /usr/local/opt/go/libexec/src/runtime/panic.go:1038 +0x215
github.com/snyk/vervet.mergePaths(0xc000400c30, 0xc000400d00, 0x1)
        /Users/gabrielenslein/go/pkg/mod/github.com/snyk/vervet@v1.2.1/merge.go:135 +0xf2
github.com/snyk/vervet.Merge(0xc000400c30, 0xc000400d00, 0x1)
        /Users/gabrielenslein/go/pkg/mod/github.com/snyk/vervet@v1.2.1/merge.go:20 +0x7f
vervet-underground/internal/storage/mem.(*Storage).mergeContentRevisions(0xc0000a0d70, {{0x14b2265, 0xc00007dab0, 0x0}, 0x0}, {0xc000402e60, 0x2, 0xc0003206e0})
        /Users/gabrielenslein/go/src/github.com/snyk/vervet/vervet-underground/internal/storage/mem/mem.go:265 +0x235
vervet-underground/internal/storage/mem.(*Storage).collateVersion(0xc0000a0d70, {{0x0, 0xed8d48000, 0x0}, 0x4})
        /Users/gabrielenslein/go/src/github.com/snyk/vervet/vervet-underground/internal/storage/mem/mem.go:235 +0x646
vervet-underground/internal/storage/mem.(*Storage).CollateVersions(0xc0000a0d70)
        /Users/gabrielenslein/go/src/github.com/snyk/vervet/vervet-underground/internal/storage/mem/mem.go:154 +0xda
vervet-underground/internal/scraper_test.TestScraperCollation(0xc0001eeb60)
        /Users/gabrielenslein/go/src/github.com/snyk/vervet/vervet-underground/internal/scraper/scraper_test.go:209 +0x805
testing.tRunner(0xc0001eeb60, 0x14c22b8)
        /usr/local/opt/go/libexec/src/testing/testing.go:1259 +0x102
created by testing.(*T).Run
        /usr/local/opt/go/libexec/src/testing/testing.go:1306 +0x35a

```
Perhaps some assistance monday?

vervet-underground/internal/storage/mem/mem.go
```diff
@@ -6,81 +6,274 @@ package mem
...
+		}
+	}
+
+	s.mu.Lock()
+	s.collatedVersions = make([]vervet.Version, len(uniqueVersions))
+	i := 0
+	for _, value := range uniqueVersions {
+		s.collatedVersions[i] = value
+		i++
+	}
```
cmars, line 198: Slightly more compact:

```suggestion
	for i, value := range uniqueVersions {
		s.collatedVersions[i] = value
	}
```

vervet-underground/internal/storage/mem/mem.go
```diff
@@ -6,81 +6,274 @@ package mem
...
+// collects the latest ContentRevision, if one exists and matches the target.
+// Once all services have been searched, call mergeContentRevisions
+func (s *Storage) collateVersion(version vervet.Version) error {
+	// number of services maximum needed
+	contentRevisions := make([]ContentRevision, 0)
+
+	// preprocessing all relevant docs in byte format before combining
+	for service, versionSlice := range s.serviceVersions {
+		// If there is an exact match on versions 1-to-1
+		if revisions, ok := s.serviceVersionMappedRevisionSpecs[service][version.String()]; ok {
```
cmars, line 215: This will need to be protected by the mutex.
```suggestion
		s.mu.RLock()
		revisions, ok := s.serviceVersionMappedRevisionSpecs[service][version.String()]
		s.mu.RUnlock()
		if ok {
```

vervet-underground/internal/storage/mem/mem.go
```diff
@@ -6,81 +6,274 @@ package mem
...
+
+		// If there is a fuzzy match on version supplied at collation for this service
+		// aka execute binarySearch on this service's available versions
+		resolvedVersion, err := versionSlice.Resolve(version)
+		if err != nil {
+			log.Error().Err(err).Msgf("Could not resolve for service %s version %s", service, version.String())
+			continue
+		}
+
+		if revisions, ok := s.serviceVersionMappedRevisionSpecs[service][resolvedVersion.String()]; ok {
```
cmars, line 231: This will need a read lock as well.

vervet-underground/internal/storage/mem/mem.go
```diff
@@ -6,81 +6,274 @@ package mem
...
+		} else {
+			// TODO: evaluate whether to use replace bool or not during merging
+			vervet.Merge(dst, src, true)
+		}
+	}
+
+	// lock memory safely here
+	s.mu.Lock()
+	s.mu.Unlock()
+	s.collatedVersionedSpecs[version] = *dst
```
cmars, line 277:
```suggestion
	s.mu.Lock()
	s.collatedVersionedSpecs[version] = *dst
	s.mu.Unlock()
```

vervet-underground/internal/storage/mem/mem.go
```diff
@@ -6,81 +6,274 @@ package mem
...
+
+func (s *Storage) GetCollatedVersionSpecs() (map[string][]byte, error) {
+	s.mu.RLock()
+	defer s.mu.RUnlock()
+
+	versionSpecs := map[string][]byte{}
+	for key, value := range s.collatedVersionedSpecs {
+		json, err := value.MarshalJSON()
+		if err != nil {
+			log.Error().Err(err).Msgf("Could not Marshal Version %s", key.String())
```
cmars, line 171: Your call, but I usually let these kinds of errors bubble up to the top and log there.

vervet-underground/internal/storage/mem/mem.go
```diff
@@ -6,81 +6,274 @@ package mem
...
+//   - calls updateCollatedVersions for a slice of unique vervet.Version entries
+//   - for each unique vervet.Version, run collateVersion to create a compiled VU openapi doc
+func (s *Storage) CollateVersions() error {
+	var errs error
+	s.updateCollatedVersions()
+	for _, version := range s.collatedVersions {
+		err := s.collateVersion(version)
+		if err != nil {
+			log.Error().Err(err).Msgf("Could not Collate Version %s", version.String())
+			errs = multierr.Append(errs, err)
```
cmars, line 157: Minor suggestion, but might be more compact and less chatty to 

```suggestion
			errs = multierr.Append(errs, errors.Wrapf(err, "failed to collate version %s", version.String())
```

and let the caller log the error? `multierr` should show all of them when you log it.

cmars (commented):
We're getting there! Just a few minor locking issues that can be easily addressed.

cmars:
> Perhaps some assistance monday?

Absolutely!

vervet-underground/internal/storage/mem/mem.go
genslein: `uniqueVersions` is a map. The mechanism is to use the map for creating uniqueness in the array, and then inserting all the values and presizing the array.

I could simplify the code to just use `append` instead if you'd rather but this is slightly more memory efficient

cmars (approved):
LGTM with latest revision of a version in collated versions.

PR #95, opened 2021-12-10
chore: remove unnecessary log messages

cmars:
There were a couple of log messages leftover from debugging the optic
linter.


cmars:
JFDI

PR #96, opened 2021-12-13
Feat/present compiled vu

genslein:
Initial draft exposing data via the web API 

Will be writing specs shortly

### DIFF
--- a/vervet-underground/internal/scraper/scraper.go
+++ b/vervet-underground/internal/scraper/scraper.go
@@ -8,6 +8,7 @@ import (
 	"io/ioutil"
 	"net/http"
 	"net/url"
+	"strings"
 	"time"
 
 	"github.com/pkg/errors"
@@ -23,17 +24,28 @@ type Storage interface {
 	// NotifyVersions tells the storage which versions are currently available.
 	// This is the primary mechanism by which the storage layer discovers and
 	// processes versions which are removed post-sunset.
-	NotifyVersions(ctx context.Context, name string, versions []string, scrapeTime time.Time) error
+	NotifyVersions(name string, versions []string, scrapeTime time.Time) error
+
+	// CollateVersions tells the storage to execute the compilation and
+	// update all VU-formatted specs from all services and their
+	// respective versions gathered.
+	CollateVersions() error
 
 	// HasVersion returns whether the storage has already stored the service
 	// API spec version at the given content digest.
-	HasVersion(ctx context.Context, name string, version string, digest string) (bool, error)
+	HasVersion(name string, version string, digest string) (bool, error)
 
 	// NotifyVersion tells the storage to store the given version contents at
 	// the scrapeTime. The storage implementation must detect and ignore
 	// duplicate version contents, as some services may not provide content
 	// digest headers in their responses.
-	NotifyVersion(ctx context.Context, name string, version string, contents []byte, scrapeTime time.Time) error
+	NotifyVersion(name string, version string, contents []byte, scrapeTime time.Time) error
+
+	// Versions fetches the Storage Versions compiled by VU
+	Versions() []string
+
+	// Version fetches the Storage Version spec compiled by VU
+	Version(version string) ([]byte, error)
 }
 
 // Scraper gets OpenAPI specs from a collection of services and updates storage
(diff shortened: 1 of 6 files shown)


vervet-underground/internal/scraper/scraper.go
```diff
@@ -34,6 +34,12 @@ type Storage interface {
 	// duplicate version contents, as some services may not provide content
 	// digest headers in their responses.
 	NotifyVersion(ctx context.Context, name string, version string, contents []byte, scrapeTime time.Time) error
+
+	// Versions fetches the Storage Versions compiled by VU
+	Versions() []string
```
cmars, line 39: Storage can likely fail any operation when we consider implementations other than `mem`. I'm fine with leaving that as a followup refactor though; internal packages give us that freedom.

vervet-underground/server.go
```diff
@@ -33,6 +35,14 @@ func main() {
 	}
 
 	log.Info().Msgf("services: %s", cfg.Services)
+	st := mem.New()
+	sc, err := scraper.New(cfg, st)
+	if err != nil {
+		logError(err)
+		panic("unable to load storage")
```
cmars, line 43: Something to consider here is, introducing an `App` or `Service` that does all the setup and starts/stops everything (like a supervisor, basically), returning the error early here. Then `main` can panic once on setup error, in one place. Also might be useful to run in an integration test harness.

cmars (commented):
Off to a great start here. Only thing I think is missing is a background goroutine that operates the scraper. We probably also want to scrape on first startup before we start serving, so we don't panic (in the anxiety sense) our clients with an empty list of spec versions.

vervet-underground/server.go
genslein: Yeah, I was going to start pulling things out to do setups. This was the first scaffold but early view here

vervet-underground/internal/scraper/scraper.go
genslein: Yeah, probably once we get to additional features this interface will change

vervet-underground/internal/scraper/scraper.go
```diff
@@ -98,13 +130,13 @@ func Clock(c func() time.Time) Option {
 }
 
 // Run executes the OpenAPI version scraping on all configured services.
-func (s *Scraper) Run(ctx context.Context) error {
+func (s *Scraper) Run() error {
```
cmars, line 133: Disagree with removing the context. `context.WithTimeout` or `context.WithCancel` would be the preferred way to cancel a scrape in progress (and already worked).

See https://pkg.go.dev/context#section-documentation for background on how context is typically used.

vervet-underground/internal/scraper/scraper_test.go
```diff
@@ -83,23 +83,23 @@ func TestScraper(t *testing.T) {
 	c.Assert(err, qt.IsNil)
 
 	// Cancel the scrape context after a timeout so we don't hang the test
-	ctx, cancel := context.WithTimeout(context.Background(), time.Second*10)
+	_, cancel := context.WithTimeout(context.Background(), time.Second*10)
```
cmars, line 86: This no longer does anything. Setting a timeout on the context ensured the scraper would shut down after 10s, so we don't potentially hang the test.

vervet-underground/internal/scraper/scraper.go
```diff
@@ -53,28 +64,49 @@ type service struct {
...
 		}
 	}
-	return s, nil
+	return nil
+}
+
+func scraperWithTimeout(storage Storage, timeout time.Duration) *Scraper {
+	return &Scraper{
+		storage: storage,
+		http:    &http.Client{Timeout: timeout},
```
cmars, line 107: A [scraper.HTTPClient](https://github.com/snyk/vervet/blob/cc8fe6323cb19a6905c6cffd3b1a1f36412db26b/vervet-underground/internal/scraper/scraper.go#L83) option can already be passed to `scraper.New`. That would allow configuration of the client timeout. And the context we were using should have already handled cleaning up all active client connections in progress and goroutines when it is canceled or times out.

vervet-underground/internal/scraper/scraper.go
```diff
@@ -113,41 +145,52 @@ func (s *Scraper) Run(ctx context.Context) error {
...
+		err = s.storage.NotifyVersion(svc.base, versions[i], contents, scrapeTime)
 		if err != nil {
 			return errors.WithStack(err)
 		}
 	}
 	return nil
 }
 
-func (s *Scraper) getVersions(ctx context.Context, svc service) ([]string, error) {
-	req, err := http.NewRequestWithContext(ctx, "GET", svc.url.String(), nil)
```
cmars, line 147: See https://pkg.go.dev/net/http#NewRequestWithContext, "the context controls the entire lifetime of a request and its response"

cmars (changes requested):
Please don't remove `context` from the scraper. Explanations below. Happy to discuss further.

vervet-underground/internal/scraper/scraper.go
genslein: https://pkg.go.dev/net/http#NewRequest wraps under the hood `context.Background()` and calls `NewRequestWithContext`. We don't need to handle things this way.

cmars: So the idea with passing in a context is that a cancel or timeout at a level above will cascade down to all the scraper goroutines and their clients. This allows you to shut down the service at the top level by canceling a "root" context, which will cascade down to all its children and immediately abort all requests in progress. Or, you could run a scrape with a timeout; run a scrape for up to 15 minutes, then alert and give up. You might need to detect a hang not only due to the service scrapes (which can be mitigated with a client timeout) but also storage issues.

genslein: Then `graceful-timeout` is far too small and should be done per run. Also we don't actually check `Done()` further down (see https://cs.opensource.google/go/go/+/refs/tags/go1.17.5:src/net/http/server.go;l=2734-2744)  So while I agree we need some context, this is setting a per http request timeout. 

We'd want a separate timeout from the initial one set from the top-level `Run` which is not really set up right now.

vervet-underground/internal/scraper/scraper.go
genslein: refactored

vervet-underground/internal/scraper/scraper.go
genslein: Let's chat about options here but the initial context object needs to be branched correctly downstream and the max timeout needs to be significantly larger

genslein: Will defer hardening `Context` layers after landing.

PR #97, opened 2021-12-13
chore: optimize optic docker cmdline, add debug logging

cmars:
This removes a lot of redundant volume mounts from the docker command line and adds a debug logging option to help with figuring out what vervet is actually running on `vervet lint`.


jcsackett (approved):
LGTM!

PR #98, opened 2021-12-14
feat: versionware; introduce version-based routing

cmars:
Add a multiplexing handler that dispatches requests to handlers based on
a version query parameter.

### DIFF
--- a/versionware/handler_test.go
+++ b/versionware/handler_test.go
@@ -1,0 +1,114 @@
+package versionware_test
+
+import (
+	"fmt"
+	"io/ioutil"
+	"net/http"
+	"net/http/httptest"
+	"testing"
+
+	qt "github.com/frankban/quicktest"
+
+	"github.com/snyk/vervet"
+	"github.com/snyk/vervet/versionware"
+)
+
+func ExampleHandler() {
+	h := versionware.NewHandler([]versionware.VersionHandler{{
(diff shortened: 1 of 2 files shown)


cmars:
@talgendler @jjmschofield you might find this useful

cmars:
@genslein I'm iterating on some feedback there. Starting this as an experiment, a possible lightweight approach to a version dispatch that might also support edge proxy use cases.

versionware/handler_test.go
```diff
@@ -0,0 +1,114 @@
...
+	"net/http/httptest"
+	"testing"
+
+	qt "github.com/frankban/quicktest"
+
+	"github.com/snyk/vervet"
+	"github.com/snyk/vervet/versionware"
+)
+
+func ExampleHandler() {
```
jcsackett, line 16: Where does this get used? I don't see it called in this test--looks like you directly implemented something similar in the tests below. Am I missing something?

jcsackett (approved):
LGTM, though I'm curious about the one func called out in my comment. I assume I'm missing something :)

versionware/handler_test.go
cmars: It gets run like a `Test*` function, but it also [shows up in godocs when you name them like this](https://go.dev/blog/examples). It'll be a nice way to play with the code.

PR #105, opened 2021-12-16
feat: fix optic git file source support; support sweater-comb npm pkg

cmars:
It turns out that the way OpenAPI spec files were being sourced and
compared one-at-a-time doesn't work for relative $refs at all. I ran
into this when QA'ing vervet + sweater-comb 1.0.x and introducing a
breaking change in a referenced schema file.

This copies a subdirectory out of the Git branch to a temporary
directory, so that relative file references are preserved intact. It
comes with the caveat that the resource paths declared in the Vervet
config need to contain all the relative references. However, this is a
reasonable constraint to add in my opinion. It also makes the Docker
volume mounts much, much simpler.

This also adds the option of running optic-ci as an npm-installed script on
the host, as an alternative to a docker container. This eliminates the
overhead of having to run the tool with Docker.

Drive-by: update the docker container image


genslein (approved):
I don’t have enough context in the optic changes but appreciate the go mod bump for VU

PR #107, opened 2021-12-17
Feat/cut vu release

genslein:
Preparing to publish first cut of Vervet-Underground to Docker Image Repo for deployment usage

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -6,6 +6,36 @@ defaults: &defaults
   docker:
     - image: circleci/golang:1.16-node
 
+vu_metadata: &vu_metadata
+    resource_class: small
+    working_directory: ~/vervet/vervet-underground
+
+ci_cd_image: &ci_cd_image
+  <<: *vu_metadata
+  docker:
+    - image: golang:1.17
+      auth:
+        username: _json_key
+        password: $GCLOUD_GCR_BUILDER
+
+commands:
+  gcr_auth:
+    description: Auth GCR
+    steps:
+      - setup_remote_docker
+      - run:
+          name: GCR auth
+          command: echo $GCLOUD_GCR_BUILDER | docker login -u _json_key --password-stdin
+            https://gcr.io/snyk-main
+  git_https_token:
+    description: Authenticate using provided token when fetching private repos from
+      github.com/snyk, use `go-private-modules` context
+    steps:
+      - run:
+          name: use private token for git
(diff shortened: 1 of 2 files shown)


.circleci/config.yml
```diff
@@ -4,6 +4,36 @@ defaults: &defaults
...
+          command: echo $GCLOUD_GCR_BUILDER | docker login -u _json_key --password-stdin
+            https://gcr.io/snyk-main
+  git_https_token:
+    description: Authenticate using provided token when fetching private repos from
+      github.com/snyk, use `go-private-modules` context
+    steps:
+      - run:
+          name: use private token for git
+          command: git config --global url."https://${GITHUB_PRIVATE_TOKEN}:x-oauth-basic@github.com/snyk".insteadOf
+            "https://github.com/snyk"
```
cmars, line 35: I don't think we need this.

vervet-underground/server.go
```diff
@@ -22,8 +22,10 @@ import (
 
 func main() {
 	var wait time.Duration
+	var interval time.Duration
```
cmars, line 25: Might be nice to have more consistency between the var and flag names. 

```suggestion
	var scrapeInterval time.Duration
```

and likewise for the other options.

cmars: Another thought: single-source these options and 12-factor app it in the process with https://github.com/spf13/viper?

cmars (commented):
A few thoughts, generally looks good so far.

vervet-underground/server.go
genslein: @cmars I think making the options nicer would probably go through cobra instead to make them flag OR config. That being said, I'd like to avoid mixing that in here.

.circleci/config.yml
```diff
@@ -6,6 +6,28 @@ defaults: &defaults
...
+
+commands:
+  gcr_auth:
+    description: Auth GCR
+    steps:
+      - setup_remote_docker
+      - run:
+          name: GCR auth
+          command: echo $GCLOUD_GCR_BUILDER | docker login -u _json_key --password-stdin
+            https://gcr.io/snyk-main
```
cmars, line 29: Was this line intended? Does it concatenate with the command string? Not sure how YAML deals with whitespace like this.

cmars (approved):

.circleci/config.yml
genslein: https://github.com/snyk/oauth-token-server/blob/main/.circleci/config.yml#L29-L37
Deals just fine by the looks of things here. I'd have also expected differently but 🤷  it's a copy-pasted block in a lot of our docker upload commands

PR #109, opened 2021-12-17
chore: set semantic-releaser to dry-run

cmars:
To avoid burning through more major versions, setting releaser to
--dry-run until we can trust it.

This way we can see what it would do in CI logs.


cmars:
JFDI

PR #111, opened 2021-12-20
chore: fix working dir start point dockerfile

genslein:
The working directory alone does not change to the subdirectory build in CircleCI.

Fix changing working directory to use `vervet-underground` directory appropriately.


genslein:
JDFI

PR #112, opened 2021-12-20
chore: go upgrade deps

genslein:
Upgrade golang dependencies and add fix for docker-compose to include config parameter by default.


genslein:
JDFI patching some vulnerabilities detected in `snyk test`

PR #114, opened 2022-01-05
wip: introduce k8s service testing

Todai88:
Please be mindful that this very much is a PoC / WIP. 

Instead of depending on a config file, we can use Kubernetes' go-library to
loop over services and identify viable services by label (`registry-web` for now).

Added a few TODOs too that should be addressed.


cmars (commented):
First and foremost: This is fricken awesome!

A thought: I'm wondering if we should separate concerns here a bit:
* Keep the scraper as a thing that scrapes lists of services
* Introduce a service discovery interface w/k8s implementation, as a thing which configures a Scraper interface with a list of discovered services.

That keeps options open if one wants to add extra services. Thinking of how prometheus is configured -- you can specify scrape targets directly by host:port/path, and also set up SD to import additional scrape targets from k8s and other sources automatically.

genslein:
I love this! I like the idea of allowing static vs. dynamic discovery here as a separate setting in VU. A great jumping off point to be sure though!

Issue #100, opened 2021-12-14
Implement explicit operation sunsetting


cmars:
Implement the explicit operation sunsetting required by the solution proposed in #99.

Proposal here is to introduce two extensions on _operations_ in versioned specs:

* One that marks the operation as deprecated. This is necessary when an operation needs to be sunset without being superseded by one in a newer version.
* One that indicates when the operation was deprecated. A linter rule (Optic CI) will need to assert that this cannot be back- or post-dated, it must match the date it lands.

cmars:
I think this can be done in a single extension on the operation, `x-snyk-deprecated-on`, which takes a date of the form `YYYY-mm-dd`.

Adding `x-snyk-deprecated-on` to an operation should cause requests to this endpoint to be regarded as deprecated, as if it had been followed by a newer release on the date given. Sunset calculation is determined according to this date and the stability of the operation being deprecated.

A Sweater Comb rule will also be needed to prevent shenanigans, such as back-dating a deprecation.

PR #115, opened 2022-01-06
feat: per-operation versioning and deprecation

### LINKED ISSUE #99: Operations in version specs should be additive over time
We found that carrying prior versions' paths forward and copying them into the next version to be cumbersome, brittle, and error prone.

Instead of requiring each version release to re-declare and re-enumerate all its operations, operations should be deprecated by newer versions if re-declared; otherwise, the older release should still resolve until deprecated and sunset.

# Scenario

To illustrate how this should work:

Initial version of a resource "things" is released as `things/2021-08-20/spec.yaml`, which declares:

```
paths:
  '/things/{id}':
    get: {...}
  '/things:
    get: {...}
```

Following version of things, `things/2021-09-30/spec.yaml` declares:

```
paths:
  '/things/{id}':
    delete: {...}
```

# Current interpretation

The current implementation sees each version spec as a standalone contract defining the resource. So, it basically interprets this as: "on 2021-09-30, two GET operations were removed and a DELETE operation was added". Ok, that might be precise enough for machines, but it makes things a lot harder on developers; you have to remember to copy over each operation, hopefully you get it right, and if you don't, users of the latest API at that stability will be in for an unpleasant surprise.

# Proposed interpretation

Proposal here is to change the interpretation to: "on 2021-09-30, a new DELETE operation was added". It doesn't "erase" the prior GETs; at 2021-09-30, those operations (path + method) still resolve to the 2021-08-20 versions.

# Solution

The only thing that changes here is how resource versions are compiled. The resulting output for the effective API versions:

`2021-08-20/spec.yaml`:

```
paths:
  /things:
    get: {...}
  /things/{id}:
    get: {...}
```

`2021-09-30/spec.yaml`:

```
paths:
  /things:
    get: {...}
  /things/{id}:
    get: {...}
    delete: {...}
```

# Implications

## Existing usage
Existing resources that were re-declaring their prior operations across versions will not need to change anything; they will not need to do this anymore in future releases, once upgrading to this change.

## Explicit operation sunsetting
In cases where an operation (path + method) needs to be removed, this deprecation and sunsetting process will need to be made explicit. A followup issue will propose how that should work.

## Release version
Because this is a pretty fundamental change in output, this change should be released as a major version bump (2.0.0).

cmars:
Fixes #99, ~~#100~~ (see note below).

This changes how operation (path + method) are resolved and merged
across spec versions within a stability level.

Operations are now cumulative. Unless a particular operation (path +
method) is redefined or explicitly deprecated, past releases are assumed
to carry over to newer versions.

BREAKING CHANGE: This changes the behavior of the compiler and resulting
output. It's also a breaking API change, requiring a major version bump.
We do not anticipate any breaking changes to OpenAPI consumers.

### DIFF
--- a/resource.go
+++ b/resource.go
@@ -167,7 +167,10 @@ func LoadResourceVersions(epPath string) (*ResourceVersions, error) {
 func LoadResourceVersionsFileset(specYamls []string) (*ResourceVersions, error) {
 	var resourceVersions ResourceVersions
 	var err error
-	pathReleases := map[string]VersionSlice{}
+	type operationKey struct {
+		path, operation string
+	}
+	opReleases := map[operationKey]VersionSlice{}
 
 	for i := range specYamls {
 		specYamls[i], err = filepath.Abs(specYamls[i])
@@ -188,14 +191,21 @@ func LoadResourceVersionsFileset(specYamls []string) (*ResourceVersions, error) 
 		if err != nil {
 			return nil, err
 		}
-		resourceVersions.versions = append(resourceVersions.versions, rc)
-		// Map of release versions per path
-		for path := range rc.Paths {
-			pathReleases[path] = append(pathReleases[path], rc.Version)
+		// Map release versions per operation
+		for path, pathItem := range rc.Paths {
+			for _, opName := range operationNames {
+				op := getOperationByName(pathItem, opName)
+				if op != nil {
+					op.ExtensionProps.Extensions[ExtSnykApiVersion] = rc.Version.String()
+					opKey := operationKey{path, opName}
+					opReleases[opKey] = append(opReleases[opKey], rc.Version)
+				}
+			}
 		}
+		resourceVersions.versions = append(resourceVersions.versions, rc)
 	}
 	// Sort release versions per path
-	for _, releases := range pathReleases {
+	for _, releases := range opReleases {
 		sort.Sort(releases)
 	}
 	// Sort the resources themselves by version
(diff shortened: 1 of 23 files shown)


resource.go
```diff
@@ -162,12 +172,87 @@ func LoadResourceVersions(epPath string) (*ResourceVersions, error) {
...
+	case "put":
+		path.Put = &ov.operation
+	case "trace":
+		path.Trace = &ov.operation
+	default:
+		panic(fmt.Sprintf("unsupported operation %q", ov.opName))
+	}
+}
+
+var timeNow = time.Now
```
cmars, line 240: This will be used to patch time in tests that I have yet to write...

cmars:
TODO: operation version carry-forward needs more comprehensive scenario tests

resource.go
ianlivingstone: Does Go not have a better mocking facility  👀

cmars:
Turns out there's more to this. Declaring a stability at the top-level of a spec is no longer appropriate when operations are versioned independently.

Assigning a stability to each operation makes more sense.

I think there's still some things that need sorting out around how to effectively promote a release.

resource.go
cmars: This method works pretty well with [C.Patch](https://pkg.go.dev/github.com/frankban/quicktest#C.Patch) actually.

cmars: Of course interfaces are best to mock with, but this is a stdlib global. It'd be more complexity to refactor everything to use a `type Clock interface` and mock it.

cmars:
This PR does not implement the explicit operation sunsetting in #100. Would like to address that in a followup, and keep this one focused on the move to per-operation versioning & carry-forward.

jcsackett (commented):
I need to take a longer look tomorrow but on first pass this looks pretty good. One minor thing to note is with all the refactoring we've dropped several godocs for exported methods, and this feels complicated enough that having those added or updated would be really valuable.

cmars:
Good shout on the godocs, I'll fix that right up!

cmars:
@jcsackett Added a couple of godoc comments missing in `spec.go`. Also realized `internal/compiler/Compiler.Build` can become simpler, since `SpecVersions` is now doing the heavy lifting of resolving all the effective release versions ahead of time.

jcsackett (approved):
This looks good to me, @cmars. Simplifying compile was a good update.

PR #118, opened 2022-01-14
fix: authenticate to Github properly when pushing release tag

cmars:


cmars:
Going to JFDI this one

PR #129, opened 2022-01-21
fix: disallow requesting API version dates in the future

cmars:
Respond 400 when a version date is requested from the future.

### DIFF
--- a/versionware/validator.go
+++ b/versionware/validator.go
@@ -5,6 +5,7 @@ import (
 	"net/http"
 	"net/url"
 	"sort"
+	"time"
 
 	"github.com/getkin/kin-openapi/openapi3"
 	"github.com/getkin/kin-openapi/openapi3filter"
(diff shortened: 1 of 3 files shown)


versionware/validator.go
```diff
@@ -118,6 +132,11 @@ func (v *Validator) Middleware(h http.Handler) http.Handler {
 			v.errFunc(w, req, http.StatusBadRequest, err)
 			return
 		}
+		if today := v.today(); requested.Date.After(today) {
```
genslein, line 135: This is kind of awkward since this doesn't update the `Validator v.today` variable but instead sets this to a local variable `today`. I'm not really sure why you'd want to save in the above if we do the calculation every time?

cmars: Poor choice of variable name here, updated and refined the comment on ValidatorConfig.Today. We definitely want to compare the current date from a clock (defaulting to the "wall clock") at the time of the request, on each request though.

genslein (approved):
mostly a nitpick: I think I'm still not sure why this isn't just a private member function of the `Validator` struct to begin with instead of exposing it. I don't think a user would want to change the definition of `Today`. Not blocking the PR on this

cmars:
Only good reason I can think of is fixing the clock in tests -- so I won't expose Today as a config option, I'll just make it patchable in this package's tests.

PR #143, opened 2022-02-08
chore: update test fixture to use sweater comb common models

cmars:


ianlivingstone (approved):

cmars:
@ianlivingstone, slight change made since your review. I found that I needed to register `format: url` in order to support the [LocationHeader](https://github.com/snyk/sweater-comb/blob/39b0f76827025ceb55c698f256f8fbf1d816bd74/components/headers/headers.yaml#L73) component out of Sweater Comb.

PR #146, opened 2022-02-18
feat: support passing extra args, ci-context to optic-ci

cmars:
Add support for passing extra args to optic-ci, as well as setting
--ci-context if the file exists. This allows optic-ci to upload the API
comparison source and results to Optic Cloud when integration with the
SaaS is desired.

Drive-by: update versionware example go.mod, it needs the same
temporary replacement as introduced in #144 in order to build.


jcsackett (approved):
LGTM

PR #148, opened 2022-02-18
feat: adds vervet's version info to cli

jcsackett:
urfave provides version info automatically as part of the cli app setup,
provided the cli is given "Version" as part of its setup.

This sets that field based on a global var in the `cmd` package, which can be
set via the ldflags arg at build time, e.g. to set the version based on git
tag as done in the Makefile.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -4,6 +4,7 @@
 .idea
 .dccache
 .vscode
+/cmd/generate_version_init.go
 
 **/server
 config.json
--- a/Makefile
+++ b/Makefile
@@ -4,7 +4,7 @@ all: lint test build
 
 .PHONY: build
 build:
-	go build -ldflags "-X 'github.com/snyk/vervet/cmd.VervetVersion=`git tag --sort=-version:refname | head -n 1`'" -a -o vervet ./cmd/vervet
+	go build -a -o vervet ./cmd/vervet
 
 # Run go mod tidy yourself
 
--- a/Makefile
+++ b/Makefile
@@ -4,7 +4,7 @@ all: lint test build
 
 .PHONY: build
 build:
-	go build -a -o vervet ./cmd/vervet
+	go build -ldflags "-X 'github.com/snyk/vervet/cmd.VervetVersion=`git tag --sort=-version:refname | head -n 1`'" -a -o vervet ./cmd/vervet
 
 # Run go mod tidy yourself
 
--- a/cmd/cmd.go
+++ b/cmd/cmd.go
@@ -15,6 +15,10 @@ import (
 	"github.com/urfave/cli/v2"
 )
 
+// VervetVersion contains vervet's version, set by ldflags at build time:
+// go build -ldflags "-X 'github.com/snyk/vervet/cmd.VervetVersion=$VERSION'"
+var VervetVersion = "0.0.1"
+
 // VervetParams contains configuration parameters for the Vervet CLI application.
 type VervetParams struct {
 	Stdin  io.ReadCloser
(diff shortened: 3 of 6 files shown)


Makefile
```diff
@@ -4,7 +4,7 @@ all: lint test build
 
 .PHONY: build
 build:
-	go build -a -o vervet ./cmd/vervet
+	go build -ldflags "-X 'github.com/snyk/vervet/cmd.VervetVersion=`git tag --sort=-version:refname | head -n 1`'" -a -o vervet ./cmd/vervet
```
cmars, line 7: I'd suggest a `go generate` to produce a `func init() {...}` that sets the version variable. `make build` doesn't actually compile the vervet that's released -- it's mainly a convenience for local development.

You could add that `go generate` right after [this line](https://github.com/snyk/vervet/blob/232a7568b90b3d8a01d27aeca7d6b88ddad7f213/.circleci/config.yml#L103) in the CI release, making use of the VERSION environment variable that gets derived from ccv.

cmd/cmd.go
```diff
@@ -15,6 +15,10 @@ import (
 	"github.com/urfave/cli/v2"
 )
 
+// VervetVersion contains vervet's version, set by ldflags at build time:
+// go build -ldflags "-X 'github.com/snyk/vervet/cmd.VervetVersion=$VERSION'"
+var VervetVersion = "0.0.1"
```
cmars, line 20: Consider leaving this empty, or an invalid value like "development", which screams "this binary isn't one of our releases".

cmd/cmd.go
```diff
@@ -67,6 +71,7 @@ func NewApp(vp VervetParams) *VervetApp {
 			Reader:    vp.Stdin,
 			Writer:    vp.Stdout,
 			ErrWriter: vp.Stderr,
+			Version:   VervetVersion,
```
cmars, line 74: Designing for test, nice!

Makefile
jcsackett: I was actually just updating the dist.bash and release.bash files too--just hadn't pushed that up yet because bash scripting and escapes are a pain 😝

scripts/dist.bash
```diff
@@ -17,6 +17,8 @@ rm -rf dist
 
 mkdir -p ./dist/bin
 
+go generate ./cmd/...
```
cmars, line 20: Consider also doing this in `scripts/release.bash`. Right now, we run this script prior in CircleCI, but in case we separate their execution later on, it won't mysteriously break.

.gitignore
```diff
@@ -4,6 +4,7 @@
 .idea
 .dccache
 .vscode
+/cmd/vervetVersion.go
```
cmars, line 7: `generated_version_init.go` might be a little clearer indication of what this is, in the event a human ever sees it.

cmars (approved):
LGTM as-is, a couple of minor suggestions below.

scripts/genversion.bash
```diff
@@ -0,0 +1,15 @@
...
+
+[ -n "${VERSION}" ]
+
+cat << EOF > cmd/vervetVersion.go
+// THIS IS A GENERATED FILE. DO NOT EDIT.
+
+package cmd
+
+func init() {
+    Vervet.App.Version = "${VERSION}"
```
genslein, line 13: I'm surprised we can't drive this off of the `go.mod` for package building. I release we used the versioning pipeline though for semver and if there's no other way, seems like a reasonable workaround.

genslein (approved):

PR #150, opened 2022-02-23
fix: upgrade kin-openapi to 0.90.0

cmars:
This updates vervet with the latest kin-openapi, which includes HTTP URI
$ref caching (https://github.com/getkin/kin-openapi/pull/469).

It also drops the replace go.mod directive, which should restore `go
install`ability of vervet.


genslein (approved):
🙌  finally can drop the replace

PR #151, opened 2022-02-24
fix: select correct architecture

eliotstocker:
* switch name of output binaries to match node arch convention

* add linux arm64 build

* use process.arch to select correct bin

### DIFF
--- a/packaging/npm/passthrough.js
+++ b/packaging/npm/passthrough.js
@@ -3,7 +3,7 @@ var path = require('path');
 
 // os and arch restrictions are handled by the package.json
 var os = process.platform;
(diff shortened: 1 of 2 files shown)


cmars:
Thanks for this! Going to QA this locally, just to be sure the filename change doesn't disrupt releases, but it seems fine!

cmars (approved):
Thanks for this! QA of npm package dist on amd64 seems fine (was mainly concerned that the rename to x64 didn't cause any disruption).

Please sign the CLA (https://cla-assistant.io/snyk/vervet?pullRequest=151) so I can land & cut a release. Thanks again!

PR #152, opened 2022-02-28
feat!: vervet generate command

cmars:
Instead of chaining vervet's templating capabilities on to OpenAPI
version aggregation and compilation, this breaks generators out to a
separate command a la `go generate`. Generators are more focused on code
generation driven by trees of OpenAPI, which "own" the generated files.

Prior experimental functionality around OpenAPI bootstrapping and
scaffolds are also removed. We want `optic run` scripting to manage
OpenAPI, and template projects instead of project generators.


cmd/cmd.go
```diff
@@ -151,34 +145,13 @@ func NewApp(vp VervetParams) *VervetApp {
...
 				}, {
-					Name:      "list",
-					Usage:     "List resource versions in a vervet project",
+					Name:      "operations",
+					Aliases:   []string{"op", "ops"},
+					Usage:     "List versioned resource operations in a vervet project",
 					ArgsUsage: "[api [resource]]",
-					Action:    VersionList,
-				}, {
-					Name:      "new",
```
genslein, line 161: Was dropping new support intentional?

cmars: Yes. `vervet generate` is intended to replace it. `vervet version new` operated in the context of a "Vervet project" owning the generators and applying them unconditionally to all resource versions. `vervet generate` breaks this functionality up so that you can generate from templates defined outside of the project.

This makes the generation capability more open and extensible for use outside of the context of a particular project -- templates can then be bundled and applied to any project. https://github.com/snyk/rest-node-generate will be one of these, planned `rest-go-generate` is another.

Also want to kill the `vervet version` command. Lots of feedback from the last workshop that it was confusing (it doesn't tell me the version of vervet?).

cmd/cmd.go
```diff
@@ -9,7 +9,6 @@ import (
 	"os"
 	"path/filepath"
 	"strings"
-	"time"
 
 	"github.com/manifoldco/promptui"
```
jcsackett, line 13: Pretty sure we don't need this anymore--it's uses in ~~version~~resource aren't there anymore, and scaffold is gone as well.

cmd/resource.go
```diff
@@ -134,101 +131,3 @@ func VersionFiles(ctx *cli.Context) error {
...
-			apis[i] = k
-			i++
-		}
-		apiName, err = v.Params.Prompt.Select("API for new version?", apis)
-		if err != nil {
-			return err
-		}
-	}
-	if resourceName == "" {
-		resourceName, err = v.Params.Prompt.Entry("Resource name?")
```
jcsackett, line 186: 1) I'm sad to see some of my first work here go away, even if it's for all the right reasons. 😆 

2) As noted above, I think this  is the bulk of the Prompt usage.

cmd/cmd.go
```diff
@@ -85,23 +76,25 @@ func NewApp(vp VervetParams) *VervetApp {
...
+						Aliases: []string{"g", "gen", "generator"},
+						Usage:   "Generators definition file",
+					},
+				},
+				Action: Generate,
 			}, {
-				Name:      "compile",
-				Usage:     "Compile versioned resources into versioned OpenAPI specs",
+				Name:      "build",
+				Usage:     "Build versioned resources into versioned OpenAPI specs",
```
jcsackett, line 97: That's a lot better than compile, actually. 👍🏼

cmd/cmd.go
```diff
@@ -151,34 +145,13 @@ func NewApp(vp VervetParams) *VervetApp {
 					Name:      "files",
 					Usage:     "List resource spec files in a vervet project",
 					ArgsUsage: "[api [resource]]",
-					Action:    VersionFiles,
+					Action:    ResourceFiles,
 				}, {
-					Name:      "list",
-					Usage:     "List resource versions in a vervet project",
+					Name:      "operations",
```
jcsackett, line 150: Is "operations" actually clearer? This might make sense as "versions" and "list resource versions", since in that context I think versions might be clear? If I'm exploring from the CLI "operations" doesn't really mean anything, does it?

internal/generator/resources.go
```diff
@@ -0,0 +1,121 @@
...
+
+// ResourceMap defines a mapping from API resource identity to its versions.
+type ResourceMap map[ResourceKey]*vervet.ResourceVersions
+
+// OperationMap defines a mapping from operation name to all versions
+// of that operation within a resource.
+type OperationMap map[string][]OperationVersion
+
+// OperationVersion represents a version of an operation, also known as an
+// endpoint.
```
jcsackett, line 28: I'm concerned the vervet terminology is drifting from what anyone writing or thinking about APIs in common usage are going to understand--I don't think "versioned operation" == "endpoint" really parses, especially given that our usage is that version is a query parameter to an endpoint, and so we have multiple definitions of one endpoint.

Or, put another way, it feels like "endpoint" contains "version",  rather than being a product of "version".

jcsackett (approved):
I've got some concerns about commands and wording, but I'm not going to block on it--I'm not at all sure that I'm correct, but I do feel like we've not quite gotten to the right terminology for end users.

cmd/cmd.go
cmars: I'm removing the scaffold command, but I'm not completely decided about removing the internals behind it. I'm trying to take advantage of the new major release version 4.x to redefine it later if we bring it back... If I leave it in 4.x, I have to commit to it.

cmd/resource.go
cmars: I failed you here in asking for the wrong thing. I'm not entirely sure we won't need prompting though. I'm hesitant to remove it entirely, this is more about trimming the public API / CLI surface to recommit to the right things in 4.x.

cmd/cmd.go
cmars: The rows in the output are per resource per version per operation. Could call these `endpoints`, but `sweater-comb run` scripts also use "operation" terminology.

`vervet resource versions` or `vervet resource list-versions`
`vervet resource operations` or `vervet resource list-operations` or `vervet resource version operations`

...

Trying to decide what we'll have to live with for 4.x... :)

cmd/cmd.go
genslein: Make sense

internal/generator/resources.go
cmars: Fair point, I think I'll stop inventing overloaded terms here and rethink this.

cmars: Going with `vervet resource info`. That seems to pair more nicely with `files`.

PR #153, opened 2022-03-02
feat: Make vervet.Version take in parameters by value

jlourenc:
My main goal with this PR is to permit chaining method calls on Version objects like:
```go
vstr := MustParseVersion("v1").AddDays(2).String()
```
This is currently not possible with some of the methods (e.g. `DateString`, `String`, `Add`, `Compare`, ...) because they are defined as pointer method.

`Version` is a relatively small structure composed of an `int`, an `int64`, a `uint64` and an address. Passing it by values, in similar way that it is done for `time.Time` seems appropriate (https://github.com/golang/go/blob/40e24a942bce7b10e23a7282e673ac8a758ca378/src/time/time.go#L248).

I took the PR a step further by updating the definition of `ParseVersion` to return a `Version` instead of a pointer to it. This is a breaking change that we may want to avoid. Looking for your inputs here. However, it brings the benefits of avoiding heap allocation and of harmonizing the `Version` API.

### DIFF
--- a/cmd/resolve.go
+++ b/cmd/resolve.go
@@ -23,7 +23,7 @@ func Resolve(ctx *cli.Context) error {
 	if err != nil {
 		return err
 	}
-	specVersion, err := specVersions.At(*version)
+	specVersion, err := specVersions.At(version)
 	if err != nil {
 		return err
 	}
--- a/resource.go
+++ b/resource.go
@@ -120,7 +120,7 @@ func (e *ResourceVersions) At(vs string) (*Resource, error) {
 	}
 	for i := len(e.versions) - 1; i >= 0; i-- {
 		ev := e.versions[i].Version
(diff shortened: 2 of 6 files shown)


cmars (commented):
I like the builder API pattern this enables. Like you point out, it does involve a breaking change to the API. Fortunately I have a PR in progress that will require a major version bump due to command line changes (#152) so you're in luck -- this can come along for the ride to v4. I'm going to place on hold pending that, but otherwise I think it's fine.

cmars:
I've changed the target branch to `staging-v4`, where I'll land all the other pending work targeting v4. Once everything lands there, I plan to land this into `main` in a single PR, triggering a major version bump.

@jlourenc Can you [sign the CLA](https://cla-assistant.io/snyk/vervet?pullRequest=153)?

jlourenc:
Sounds good. That means if I want to introduce other breaking changes, now is the time ? 😄 

CLA signed.

PR #157, opened 2022-03-04
feat: optic-ci lint exceptions

cmars:
Allow making very narrow exceptions for specs that do not pass optic-ci
lint rules as they currently are.

Exceptions are well, exceptional, but might need to be made to unblock
work that can't wait for rules to catch up. Exceptions are made based
on the spec file's relative path in the project, and its sha256. So any
subsequent change in the file will not fall under the exception.

### DIFF
--- a/internal/linter/optic/linter.go
+++ b/internal/linter/optic/linter.go
@@ -4,6 +4,8 @@ package optic
 import (
 	"bufio"
 	"context"
+	"crypto/sha256"
+	"encoding/hex"
 	"encoding/json"
 	"fmt"
 	"io"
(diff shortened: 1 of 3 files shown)


internal/linter/optic/linter.go
```diff
@@ -178,7 +182,9 @@ func (o *Optic) Run(ctx context.Context, root string, paths ...string) error {
 	}
 	for i := range paths {
 		comparison, volumeArgs, err := o.newComparison(paths[i], fromFilter, toFilter)
-		if err != nil {
+		if err == errHasException {
```
gablaxian, line 185: This is where my lack of Go experience fails me. It's weird that `errHasException` is defined much lower down in the script and is actually what looks to be an error string object? `fmt.Errorf("file is skipped due to lint exception")`

Is that right?

cmars: `fmt.Errorf` returns an [error](https://pkg.go.dev/builtin#error). `errHasException` is a global that's assigned during module init, guaranteed to be initialized before this code runs.

gablaxian (approved):

PR #159, opened 2022-03-07
feat: custom generator functions in JS, test component templating

cmars:
Templating over the kin-openapi/openapi3 document object model is
generally powerful enough to enumerate types in various languages, as
the test in this PR demonstrates for Typescript.

However, it's inevitable that such templates will need language-specific
adaptation functions. Rather than try to include all those batteries in
the provided builtin template functions, this PR adds support for
generator function extension with Javascript.

Note that goja is the same embedded JS engine used by Grafana k6.

### DIFF
--- a/config/generator.go
+++ b/config/generator.go
@@ -7,12 +7,12 @@ type Generators map[string]*Generator
 
 // Generator describes how files are generated for a resource.
 type Generator struct {
-	Name     string                    `json:"-"`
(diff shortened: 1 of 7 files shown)


garyposter:
😮 This is super cool!!!!!!!!!

(The test would be even cooler if it showed the integer -> number thing working 😛 )

PR #163, opened 2022-03-14
feat: conflict-free component relocation during merge

cmars:
When merging OpenAPI documents, detect colliding component definitions
and relocate them to non-conflicting component names, updating
references accordingly.

This "scopes" component references to the original version spec in which
they were originally referenced.

[WIP: still working through some test cases, issues with overlays]

### DIFF
--- a/deduplicate.go
+++ b/deduplicate.go
@@ -1,0 +1,169 @@
+package vervet
+
+import (
+	"crypto/sha256"
+	"encoding/hex"
+	"encoding/json"
+	"fmt"
+	"sort"
+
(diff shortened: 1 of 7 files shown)


cmars:
This is a terrible idea for several reasons:
- It encourages and embraces model drift.
- A lot of complexity hidden in relocating components that reference components.

It's far better to error with a clear message indicating which components are in conflict.

Renaming a component is very easy for a developer to do. It's not a breaking change in an API. It just forces an intentional distinct reference.

It will be far easier to do ask for a rename on occasion, than untangle a tangled web of not-quite-identical components we allow by auto-renaming.

PR #164, opened 2022-03-16
feat: collate resource versions with conflict detection

cmars:
Error on conflicting component names when merging OpenAPI specs. Error
message indicates which component is in conflict, and in which files.

In #163, automatic renaming of conflicts was explored and rejected due
to complexity and potential negative impacts to API clarity.

Test fixtures are updated because they had a conflict that was silently
undetected.


collator.go
```diff
@@ -0,0 +1,199 @@
...
+// nil.
+func (c *Collator) Result() *openapi3.T {
+	return c.result
+}
+
+// Collate merges a resource version into the current result.
+func (c *Collator) Collate(rv *ResourceVersion) error {
+	if c.result == nil {
+		c.result = &openapi3.T{}
+	}
```
jcsackett, line 38: Is it worthwhile to initialize it here, and not just set it to this in `NewCollator`? Is there a use case I'm missing where a collator is being used w/o Collate where the result being nil is useful?

jcsackett (approved):
LGTM, with one question (that's not a blocker either way).

genslein:
@cmars, is the collator going to be reused in VU?

cmars:
It could be! We might need to refactor it some when we take it there, I have a vague feeling it will need hooks in VU to get richer content out than just errors for feeding back into deploy. Something like [cmp.Reporter](https://pkg.go.dev/github.com/google/go-cmp/cmp#Reporter), for collation.

PR #165, opened 2022-03-16
fix: tag conflicts, error reporting in collator

cmars:
Tags should be allowed to be duplicated among resource versions and
specs -- this is how multiple operations share a tag. This change allows
that to happen.

Tag conflict errors are returned when tags having the same name have
a different description or extensions. This prevents one resource
version from mysteriously overwriting that of another. The conflicting
files are referenced clearly in the error message.

Finally, all conflict errors found in the entire collate process are
returned. This avoids a "pulling teeth" interaction where the user fixes
one problem just to find the next one.

### DIFF
--- a/collator.go
+++ b/collator.go
@@ -15,6 +15,7 @@ type Collator struct {
 	result           *openapi3.T
 	componentSources map[string]string
 	pathSources      map[string]string
+	tagSources       map[string]string
 }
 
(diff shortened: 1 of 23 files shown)


jcsackett (commented):
Looks good; really appreciate that it's finding all errors.

jcsackett (approved):

PR #171, opened 2022-04-05
fix: minor improvements to versionware pkg

jlourenc:
Minor fixes in `versionware` package:
- `fmt.Errorf` replaces by `errors.New` when no formatting is needed
- `map` instantiation with pre-defined size
- `for` loops using value assignment
- removed duplicated server URL override for each OpenAPI doc

### DIFF
--- a/versionware/validator.go
+++ b/versionware/validator.go
@@ -1,6 +1,7 @@
 package versionware
 
 import (
+	"errors"
 	"fmt"
 	"net/http"
 	"net/url"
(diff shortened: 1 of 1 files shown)


versionware/validator.go
```diff
@@ -79,13 +80,10 @@ func NewValidator(config *ValidatorConfig, docs ...*openapi3.T) (*Validator, err
...
 		case "http", "https":
 		case "":
-			return nil, fmt.Errorf("invalid ServerURL: missing scheme")
+			return nil, errors.New("invalid ServerURL: missing scheme")
 		default:
 			return nil, fmt.Errorf("invalid ServerURL: unsupported scheme %q (did you forget to specify the scheme://?)", serverURL.Scheme)
 		}
-		for i := range docs {
-			docs[i].Servers = []*openapi3.Server{{URL: serverURL.String()}}
-		}
```
cmars, line 88: I think removing this might break users of this package?

cmars (commented):
Just one concern below -- why is the Server.URL override removed?

versionware/validator.go
cmars: Or if it's duplicated... where?

jlourenc: A few lines below: https://github.com/snyk/vervet/blob/2b0a26aa4f08449ce6e061631ff6931dd205ffd0/versionware/validator.go#L101

jlourenc: The following test validates that override still occurs:
https://github.com/snyk/vervet/blob/2b0a26aa4f08449ce6e061631ff6931dd205ffd0/versionware/validator_test.go#L500-L503

cmars: Indeed, good catch!

cmars (approved):
Looks amazing, thanks for this!

PR #173, opened 2022-04-07
feat: support for multiple duplicate build output paths per API

cmars:
Why?

Some services have a need to write compiled OpenAPI specs to multiple
destinations. This tends to happen in Typescript services in which we
want to write compiled OpenAPI specs to two places:

- src/**, for commiting into git and ingesting into Backstage
- dist/**, for runtime module access to compiled specs

We could just npm script this in these services, but when a project has
several APIs, this can become fragile -- you now have to keep
file-copying-scripts in sync with Vervet config. Miss something, and what
worked in local dev breaks in deployments.

To simplify this and reduce the error surface, Vervet can instead allow
declaring multiple output paths per API.

### DIFF
--- a/config/api.go
+++ b/config/api.go
@@ -70,8 +70,9 @@ type Overlay struct {
 // Output defines where the aggregate versioned OpenAPI specs should be created
 // during compilation.
 type Output struct {
-	Path   string `json:"path"`
-	Linter string `json:"linter"`
+	Path   string   `json:"path,omitempty"`
+	Paths  []string `json:"paths,omitempty"`
+	Linter string   `json:"linter"`
 }
 
 func (a APIs) init(p *Project) error {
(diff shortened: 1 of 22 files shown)


internal/compiler/compiler_test.go
```diff
@@ -50,12 +52,43 @@ apis:
       - inline: |-
           servers:
             - url: ${API_BASE_URL}
-              description: Snyk API
+              description: Test API v3
```
mrded, line 55:
```suggestion
              description: Test API REST
```

internal/compiler/compiler_test.go
```diff
@@ -50,12 +52,43 @@ apis:
...
+      - linter: resource-rules
+        path: 'testdata/resources'
+        excludes:
+          - 'testdata/resources/schemas/**'
+    overlays:
+      - include: 'testdata/resources/include.yaml'
+      - inline: |-
+          servers:
+            - url: ${API_BASE_URL}
+              description: Test API v3
```
mrded, line 83:
```suggestion
              description: Test API REST
```

mrded (approved):

carissadean (approved):

config/api.go
```diff
@@ -108,7 +109,11 @@ func (a APIs) init(p *Project) error {
 				}
 			}
 		}
-		if api.Output != nil && api.Output.Linter != "" {
+		if api.Output != nil {
+			if len(api.Output.Paths) > 0 && api.Output.Path != "" {
+				return fmt.Errorf("output should specify one of 'path' or 'paths', not both (apis.%s.output)",
```
cmars, line 114: It now occurs to me, I could add a config test case for this...

PR #174, opened 2022-04-07
feat: generate backstage API model entities for compiled versions

cmars:
If a project contains a Backstage catalog-info.yaml:
- Create/replace API definitions for Vervet compiled API versions
  See
  https://backstage.io/docs/features/software-catalog/descriptor-format#kind-api
  for background on what these are and contain.
- Keep all other developer-maintained content in the file intact,
  including comments, documents that Vervet doesn't recognize or have
  any business messing with.

This is code generation that cooperates with human-authored content in a
multi-document YAML file. Vervet owns the generated API docs, and only
creates/replaces/removes those to reflect the current state of the
service's API versions.

Why?

For services versioning their APIs with Vervet, releases will
automatically show up in Backstage. Vervet is a natural place to do
this, since it builds all the API versions of that service and knows
their lifecycle stage.

This PR also introduces some of the policy changes we're going to soon
introduce for experimental stability versions:
- Experimental releases are not intended to be used long term, but
  either promoted or sunset. So,
- Experimental releases are deprecated upon release, and automatically
  considered sunset after 90 days.


cmars:
@love-bhardwaj You might be interested in this one as well, from a Backstage integration / automation perspective. It's a different angle from https://github.com/cmars/greenroom but there are some YAML-wrangling patterns here we might make use of.

love-bhardwaj:
> @love-bhardwaj You might be interested in this one as well, from a Backstage integration/automation perspective. It's a different angle from https://github.com/cmars/greenroom but there are some YAML-wrangling patterns here we might make use of.

@cmars this looks awesome as well. I am happy to go with any of the two, whichever works the best. We can discuss more about it in our kickoff meeting but I am excited 🚀

internal/backstage/backstage.go
```diff
@@ -0,0 +1,288 @@
...
+}
+
+// findNode returns the first node matching the given path of element keys
+// under a 'root' node.
+//
+// This function is a really simple and naive sort of path expression resolver
+// good enough for the YAML manipulation needs of this backstage integration.
+//
+// It may not be suitable for general purpose YAML navigation.
+func findNode(root *yaml.Node, path ...string) *yaml.Node {
```
jcsackett, line 247: Does the go yaml lib not provide a way to navigate yaml? That seems like an oversight on its part.

Not blocking on implementing our own version, it just feels like a strange thing to have to implement. May be the python programmer in me protesting, though 😝

internal/backstage/backstage_test.go
```diff
@@ -0,0 +1,267 @@
...
+    snyk.io/vervet/version/date: "2021-08-20"
+    snyk.io/vervet/version/lifecycle: sunset
+    snyk.io/vervet/version/stability: experimental
+spec:
+  type: openapi
+  lifecycle: sunset
+  owner: someone-else
+  definition:
+    $text: output/2021-08-20~experimental/spec.json
+`[1:]
```
jcsackett, line 239: Is this something that makes more sense to have as a testdata fixture? It's a really long block of text in the middle of the test file--that said, it's not actually a problem, so if you want to leave it here rather than set up loading fixtures, I suppose that's reasonable too. 🤷🏼

resource.go
```diff
@@ -17,12 +17,18 @@ const (
 	// spec with its API release stability level.
 	ExtSnykApiStability = "x-snyk-api-stability"
 
+	// ExtSnykApiLifecycle is used to annotate compiled OpenAPI with lifecycle
+	// stage: releases, deprecated or sunset. It is applied at the top-level as
+	// well as per-operation.
+	ExtSnykApiLifecycle = "x-snyk-api-lifecycle"
```
jcsackett, line 23: Don't we have headers for sunset and deprecation already? Or is this just to make it more explicit without one having to extrapolate from something being marked as deprecated by another version? I'm ok with that since this is all auto generated so being out of sync isn't a concern, but we've also said before we want to minimize using extensions too much, so figure it's worth raising as a question here.

version.go
```diff
@@ -168,7 +168,7 @@ const (
 	SunsetWIP = 0
 
 	// SunsetExperimental is the duration past deprecation after which an experimental version may be sunset.
-	SunsetExperimental = 31 * 24 * time.Hour
+	SunsetExperimental = 24 * time.Hour
```
jcsackett, line 171: 🥳

jcsackett (approved):
LGTM. A few comments/questions, but nothing blocking. Thanks!

resource.go
cmars: Providing the extension explicitly here gives us more of a single-source of truth on a version's lifecycle state -- if we want to change lifecycle policy, we can do it in one place (vervet) rather than in each handler / middleware that needs to provide these headers, if they act as a passthrough (use the extension to determine the header).

I see these extensions as being primarily intended for libraries (like rest-go-libs, rest-node-libs) as well as docs.

PR #175, opened 2022-04-08
feat: adding initial testing for localstack s3 usage

genslein:
This is working validating with `make test-localstack`.
ref: https://aws.github.io/aws-sdk-go-v2/docs/handling-errors/

This enables using S3 as a backing persistence and retrieval store for all Vervet ContentRevisions and CollatedVersion specs.


vervet-underground/internal/storage/s3/aws_s3_client_test.go
```diff
@@ -1,17 +1,146 @@
...
+	if isCIEnabled(t) {
+		return nil, fmt.Errorf("CI disables localstack")
+	}
+
+	env := map[string]string{
+		"MAIN_CONTAINER_NAME": "localstack",
+		"EDGE_PORT":           "4566",
+		"HOST_TMP_FOLDER":     "/tmp/localstack",
+		"DOCKER_HOST":         "unix:///var/run/docker.sock",
+	}
```
cmars, line 95: I'd never assume a particular port is available in a `go test` test. Or that I have permissions to access a Docker socket. Maybe I want to use podman? Or CI connects to a remote Docker instance?

For an integration test like this, I'd orchestrate the whole thing with docker-compose, and then configure ports, URLs and such completely via env vars when running the test itself in a container. Kind of like how the sample service CI works w/envoy &tc.

With a repeatable isolated environment like that, these integration tests could be unit tests that t.Skip() unless a "this is a CI test" env var is set. That way you can `go test` some things without a heavy setup.

cmars (commented):
Just some idle thoughts, this seems early stage. Do we want to continue with the AWS-specific storage? I thought WAL-G might replace it.

vervet-underground/internal/storage/s3/aws_s3_client_test.go
genslein: See my draft comment in the top

genslein:
the `wal-g` storages repo was sunset and internalized to wal-g. WALG only exports the full executable, not the library so can't reuse it. Could ask if they would re-export it.

vervet-underground/internal/storage/s3/aws_s3_client.go
```diff
@@ -4,19 +4,41 @@ package s3
 
 import (
 	"context"
+	"fmt"
+	"github.com/aws/aws-sdk-go-v2/credentials"
```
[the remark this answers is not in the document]
genslein: see https://github.com/snyk/vervet/pull/176

genslein:
@cmars its failing on test containers go for snyk. Got magic to exclude test dependencies?

vervet-underground/internal/storage/s3/client.go
```diff
@@ -0,0 +1,128 @@
...
+		Bucket: aws.String(storage.BucketName),
+		Key:    aws.String(key),
+		ACL:    types.ObjectCannedACLPublicRead,
+		Body:   reader,
+	}
+
+	r, err := s3Client.c.PutObject(context.Background(), &p)
+	log.Printf("S3 PutObject response: %+v", r)
+	if err != nil {
+		panic(err)
```
davidharrigan, line 93: should we be returning an error here instead?

genslein: Yup changed to return err

vervet-underground/internal/storage/s3/client.go
```diff
@@ -0,0 +1,505 @@
...
+
+// GetObject nice wrapper around the S3 GetObject request.
+func (s *Storage) GetObject(key string) ([]byte, error) {
+	p := s3.GetObjectInput{
+		Bucket: aws.String(BucketName),
+		Key:    aws.String(key),
+	}
+
+	r, err := s.client.GetObject(context.Background(), &p)
+	if smith := handleAwsError(err); smith != nil {
```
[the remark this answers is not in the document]
davidharrigan: ah maybe because `github.com/aws/smithy-go`?

davidharrigan (commented):
overall, looks great. I'll come back and take another look at `CollateVersions` based on our discussions earlier.

vervet-underground/internal/storage/s3/client.go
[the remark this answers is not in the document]
genslein: Exactly, smithy-go is a special flavor of errors

vervet-underground/internal/storage/s3/client.go
```diff
@@ -0,0 +1,505 @@
...
+	}
+
+	// if the digest file exists, it counts as a match, no-op, no change
+	if len(serviceVersionRevision) > 0 {
+		return nil
+	}
+
+	// Since the digest doesn't exist, add the whole key path
+	reader := bytes.NewReader(currentRevision.Blob)
+	_, putErr := s.PutObject(key, reader)
```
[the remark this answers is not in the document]
genslein: Linter errors for shadowing vars. If there's a way around it, happy to revert

vervet-underground/internal/storage/s3/client_test.go
```diff
@@ -0,0 +1,192 @@
...
+	},
+}
+
+func TestPutObject(t *testing.T) {
+	// Arrange
+	c := qt.New(t)
+	if isCIEnabled(t) {
+		c.Assert(true, qt.IsTrue)
+		return
+	}
```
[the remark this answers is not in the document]
genslein: Until we enable the localstack tests in CI, not really sure there is

vervet-underground/internal/storage/s3/client.go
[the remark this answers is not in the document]
davidharrigan: hmm I don't see anything being shadowed here, I wonder this looked differently before?

vervet-underground/internal/storage/s3/client.go
```diff
@@ -0,0 +1,505 @@
...
+			return getErr
+		}
+
+		// Assuming version is valid in path uploads
+		parsedVersion, parseErr := vervet.ParseVersion(version)
+		if parseErr != nil {
+			log.Error().Err(parseErr).Msg("unexpected version path in S3. Validate Service Revision uploads")
+			return parseErr
+		}
+		blob, readerErr := io.ReadAll(rev.Body)
```
davidharrigan, line 206: do we need to close the body here?  Maybe lines 194-here can be extracted as a function so we can `defer` body close.

vervet-underground/internal/storage/s3/client.go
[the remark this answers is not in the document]
genslein: using `_, err := s.PutObject` or  `_, err = s.PutObject` it screamed very loudly I made the above change.

vervet-underground/internal/storage/s3/client.go
genslein: Add an immediate closer after reading for fast fail

vervet-underground/internal/storage/s3/client_test.go
```diff
@@ -0,0 +1,192 @@
...
+
+	//ctx := context.Background()
+	//ctr, err := setupTestContainer(ctx, t)
+	//defer teardownTestContainer(ctx, t, ctr)
+	//c.Assert(err, qt.IsNil)
```
davidharrigan, line 44: maybe mark this as todo with a short description? I'm guessing we want to resurrect this when we can use test container here?

davidharrigan (approved):
couple more small comments, but looks great! 🚀

genslein:
This is now usable in `docker-compose up -d`

I just missed demo time so next one 😭 

add a `config.json` file with the following config locally:

```json
{
  "services":
  [
    "https://api.dev.snyk.io/v3"
  ],
  "host": "0.0.0.0",
  "storage": {
    "type": "s3",
    "region": "us-east-1",
    "endpoint": "http://localstack:4566",
    "accessKey": "test",
    "secretkey": "test",
    "sessionkey":"test"
  }
}
```

vervet-underground/Makefile
```diff
@@ -2,18 +2,23 @@ GOMOD=go mod
 GOBUILD=go build
 GOTEST=go test
 
-.PHONY: run fmt lint tidy test test-coverage
+.PHONY: run fmt lint tidy test test-coverage clean test-localstack
```
cmars, line 5: If we place a PHONY next to each target (which I tend to prefer, as it's harder to keep a single line like this up to date over time) then we could/should remove this line.

vervet-underground/config/config.go
```diff
@@ -8,8 +8,9 @@ import (
 
 // ServerConfig defines the configuration options for the Vervet Underground service.
 type ServerConfig struct {
-	Host     string   `json:"host"`
-	Services []string `json:"services"`
+	Host     string            `json:"host"`
+	Services []string          `json:"services"`
+	Storage  map[string]string `json:"storage"`
```
cmars, line 13: Could this be more strongly-typed than a free-form string map? Perhaps a

```
type StorageConfig struct {
  Type string `json:"type"`
}
```

?

vervet-underground/internal/scraper/s3_scraper_test.go
```diff
@@ -0,0 +1,180 @@
...
+			log.Warn().Err(err).Msg("failed to delete Object")
+		}
+	}
+}
+
+func TestS3Scraper(t *testing.T) {
+	c := qt.New(t)
+	if isCIEnabled() {
+		c.Assert(true, qt.IsTrue)
+		return
```
cmars, line 63:
```suggestion
		c.Skip("CI not enabled")
```

https://pkg.go.dev/testing#T.Skip doesn't need the early return -- it will end the test for you.

vervet-underground/internal/storage/s3/README.md
```diff
@@ -0,0 +1,11 @@
+# S3 Local Testing
+
+Prerequisites:
+
+* LocalStack [CLI installed](https://docs.localstack.cloud/get-started/#installation)
+  * `python3 -m pip install localstack`
+* If you need to make changes
+`localstack config validate` to verify docker-compose setup
```
cmars, line 8:
```suggestion
  * `localstack config validate` to verify docker-compose setup
```

vervet-underground/internal/storage/s3/README.md
```diff
@@ -0,0 +1,11 @@
...
+
+To validate in Docker-compose `docker-compose up -d`
+* `localhost:4566/health` will show status of all mimicked services
```
cmars, line 11:
```suggestion
* To validate in Docker-compose `docker-compose up -d`
  * `localhost:4566/health` will show status of all mimicked services
```

vervet-underground/internal/storage/s3/client.go
```diff
@@ -0,0 +1,517 @@
...
+	"github.com/getkin/kin-openapi/openapi3"
+	"github.com/rs/zerolog/log"
+	"github.com/snyk/vervet/v4"
+	"go.uber.org/multierr"
+
+	"vervet-underground/internal/storage"
+)
+
+const (
+	BucketName             = "vervet-underground-specs"
```
cmars, line 29: Is this something we might want to make configurable? Perhaps this should be defaultBucketName.

Do these consts need to be exported outside this package?

vervet-underground/internal/storage/s3/client.go
```diff
@@ -0,0 +1,517 @@
...
+	CollatedVersionsFolder = "collated-versions/"
+	ServiceVersionsFolder  = "service-versions/"
+)
+
+type StaticKeyCredentials struct {
```
cmars, line 34: Please godoc exported symbols. Or, perhaps we don't need to export them outside `s3`?

vervet-underground/internal/storage/s3/client.go
```diff
@@ -0,0 +1,517 @@
...
+	client *s3.Client
+
+	// TODO: Implement read-through memory caching of collated Specs
+	// Cached collated specs for serving traffic
+	//mu                     sync.RWMutex
+	//collatedVersions       vervet.VersionSlice
+	//collatedVersionedSpecs storage.CollatedVersionMappedSpecs
+}
+
+func New(awsCfg *Config) *Storage {
```
cmars, line 56: Can this return the `Storage` interface, hiding the s3 impl completely?

vervet-underground/internal/storage/s3/client.go
```diff
@@ -0,0 +1,517 @@
...
+
+	// Create the resource client
+	s3Client := s3.NewFromConfig(awsCfgLoader, func(o *s3.Options) {
+		o.UsePathStyle = true
+	})
+	st := &Storage{client: s3Client}
+	err = st.CreateBucket()
+	if err != nil {
+		panic(err)
+	}
```
cmars, line 102: `New` should return a (Storage, error) and propagate the error rather than panic.

vervet-underground/internal/storage/s3/client.go
```diff
@@ -0,0 +1,517 @@
...
+	awsCfgLoader, err := config.LoadDefaultConfig(context.TODO(),
+		config.WithRegion(awsCfg.AwsRegion),
+		config.WithCredentialsProvider(credentials.NewStaticCredentialsProvider(
+			awsCfg.Credentials.AccessKey,
+			awsCfg.Credentials.SecretKey,
+			awsCfg.Credentials.SessionKey)),
+		config.WithEndpointResolverWithOptions(customResolver),
+	)
+	if err != nil {
+		log.Fatal().Err(err).Msg("Cannot load the AWS configs")
```
cmars, line 91: Error should be returned, possibly with annotation. Even if that results in a process exit nonzero, we should handle that at the top in service initialization.

cmars (changes requested):
Some comments below.

Recommended changes on docs are just nits.

Things I'd like addressed:

- Errors should be propagated and returned rather than panicked on.
- Naming of error variables. Not using `err` tends to be... error prone, and harder to read. If a linter rule is getting in the way and not letting you reuse `err`, I'm seriously wondering if it is not a good rule and should be removed from our golangci-lint config. Open to discussion here, but I'm really going to need to understand the reason behind it...

vervet-underground/internal/storage/s3/client.go
```diff
@@ -0,0 +1,517 @@
...
+}
+
+type Storage struct {
+	client *s3.Client
+
+	// TODO: Implement read-through memory caching of collated Specs
+	// Cached collated specs for serving traffic
+	//mu                     sync.RWMutex
+	//collatedVersions       vervet.VersionSlice
+	//collatedVersionedSpecs storage.CollatedVersionMappedSpecs
```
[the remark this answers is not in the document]
genslein: Sure

vervet-underground/config/config.go
genslein: Each config for S3 and GCS has slightly different config requirements. That means additional JsonParsing in each Storage Client. I'd rather save that hardening for after this PR and #184.

vervet-underground/internal/storage/s3/client.go
genslein: I could move them to top-level storage. Ideally, users wouldn’t have to guess to create a bucket and folder structure to use here

vervet-underground/internal/storage/storage.go
```diff
@@ -0,0 +1,56 @@
...
+
+// CollatedVersionMappedSpecs Compiled aggregated spec for all services at that given version.
+type CollatedVersionMappedSpecs map[vervet.Version]openapi3.T
+
+func GetSantizedHost(name string) string {
+	host := name
+	if strings.HasPrefix(name, "http") {
+		parsedUrl, err := url.Parse(name)
+		if err != nil {
+			log.Warn().Err(err).Msgf("service.base url misconfigured. Falling back %v", name)
```
cmars, line 51: If err != nil, I think we may end up with a nil deref below. parsedUrl would be nil...

vervet-underground/server.go
```diff
@@ -46,7 +48,12 @@ func main() {
 
 	// initialize Scraper
 	ticker := time.NewTicker(scrapeInterval)
-	st := mem.New()
+	st, err := initializeStorage(cfg)
+	if err != nil {
+		logError(err)
+		panic("unable to initialize storage client")
```
cmars, line 54: Consider using https://pkg.go.dev/github.com/rs/zerolog?utm_source=godoc#Logger.Fatal ? Though, a panic will definitely tell you where it happened...

vervet-underground/server.go
```diff
@@ -151,8 +158,8 @@ func healthHandler(router *mux.Router, services []string) {
 		w.WriteHeader(http.StatusOK)
 		w.Header().Set("Content-Type", "application/json")
 		encoder := json.NewEncoder(w)
-		encodeErr := encoder.Encode(map[string]interface{}{"msg": "success", "services": services})
-		if encodeErr != nil {
+		err := encoder.Encode(map[string]interface{}{"msg": "success", "services": services})
+		if err != nil {
```
cmars, line 162: nit:
```suggestion
		if err := encoder.Encode(map[string]interface{}{"msg": "success", "services": services}); err != nil {
```

cmars (commented):
A few minor things need fixing, on closer review, but I think we're getting close!
* Some places look like a nil deref is possible on errors
* Test cleanup concerns

vervet-underground/internal/storage/s3/client.go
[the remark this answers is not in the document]
genslein: resolved, removed bugged ci rule

vervet-underground/internal/storage/s3/client.go
genslein: Done

vervet-underground/internal/scraper/scraper.go
```diff
@@ -87,7 +56,11 @@ func setupScraper(s *Scraper, cfg *config.ServerConfig, options []Option) error
 		if err != nil {
 			return errors.Wrapf(err, "invalid service %q", cfg.Services[i])
 		}
+		// Handle for local/smaller deployments and tests
 		s.services[i] = service{base: cfg.Services[i], url: u}
+		if u.Hostname() == "localhost" || u.Hostname() == "127.0.0.1" {
```
cmars, line 61: Why is this special case necessary?

vervet-underground/internal/storage/s3/client.go
```diff
@@ -0,0 +1,503 @@
...
+
+		// Assuming version is valid in path uploads
+		parsedVersion, err := vervet.ParseVersion(version)
+		if err != nil {
+			log.Error().Err(err).Msg("unexpected version path in S3. Validate Service Revision uploads")
+			return err
+		}
+
+		blob, err := io.ReadAll(rev.Body)
+		err = multierr.Append(err, rev.Body.Close())
```
cmars, line 204: Interesting. Usually I'd fail on the read error, but `defer rev.Body.Close()` and not fail on that error. If we're successful in reading the entire stream, but fail to close it, did we fail in our objective here? This might be a really remote-to-impossible situation, but it's a subtle difference in behavior...

vervet-underground/internal/storage/s3/client.go
```diff
@@ -0,0 +1,503 @@
...
+func (s *Storage) PutObject(key string, reader io.Reader) (*s3.PutObjectOutput, error) {
+	p := s3.PutObjectInput{
+		Bucket: aws.String(storage.BucketName),
+		Key:    aws.String(key),
+		ACL:    types.ObjectCannedACLPublicRead,
+		Body:   reader,
+	}
+
+	r, err := s.client.PutObject(context.Background(), &p)
+	log.Printf("S3 PutObject response: %+v", r)
```
cmars, line 277: Should this be logged at debug level?

vervet-underground/internal/storage/s3/client.go
```diff
@@ -0,0 +1,503 @@
...
+func (s *Storage) PutCollatedSpecs(objects map[vervet.Version]openapi3.T) (res []s3.PutObjectOutput, smith error) {
+	res = make([]s3.PutObjectOutput, 0)
+	for key, file := range objects {
+		jsonBlob, err := file.MarshalJSON()
+		if err != nil {
+			return nil, fmt.Errorf("failure to marshal json for collation upload: %w", err)
+		}
+		reader := bytes.NewReader(jsonBlob)
+		r, err := s.PutObject(storage.CollatedVersionsFolder+key.String()+"/spec.json", reader)
+		if smith = handleAwsError(err); smith != nil {
```
cmars, line 296: Mr. Anderson? Oh, ok, I see David's question answered below :D

vervet-underground/internal/storage/s3/client.go
```diff
@@ -0,0 +1,503 @@
...
+
+// DeleteObject nice wrapper around the S3 DeleteObject request.
+func (s *Storage) DeleteObject(key string) error {
+	p := s3.DeleteObjectInput{
+		Bucket: aws.String(storage.BucketName),
+		Key:    aws.String(key),
+	}
+
+	r, err := s.client.DeleteObject(context.Background(), &p)
+	log.Printf("S3 DeleteObject response: %+v", r)
```
cmars, line 347: Debug level?

vervet-underground/internal/storage/s3/client.go
```diff
@@ -0,0 +1,503 @@
...
+		Bucket: aws.String(storage.BucketName),
+		Prefix: aws.String(key),
+	}
+
+	if delimeter != "" {
+		p.Delimiter = aws.String("/")
+	}
+
+	r, err := s.client.ListObjectsV2(context.Background(), &p)
+	log.Printf("S3 ListObject response: %+v", r)
```
cmars, line 401: Debug?

vervet-underground/internal/storage/s3/client.go
```diff
@@ -0,0 +1,503 @@
...
+	host := storage.GetSantizedHost(name)
+	return fmt.Sprintf("%v%v/%v/%v.json", storage.ServiceVersionsFolder, host, version, digest)
+}
+
+func parseServiceVersionRevisionKey(key string) (string, string, string) {
+	// digest can have "/" chars, so only split for service and version
+	arr := strings.SplitN(strings.TrimPrefix(key, storage.ServiceVersionsFolder), "/", 3)
+	if len(arr) != 3 {
+		log.Warn().Msgf("Service Content Revision not able to be parsed: %v", key)
+		return "", "", ""
```
cmars, line 460: This seems like an error condition. Generally I'd recommend returning it. Kind of like how https://pkg.go.dev/net#SplitHostPort does in the stdlib. Being internal & not exported, I'll leave it up to you, my only real ask is that you handle & test this case.

vervet-underground/internal/storage/s3/client.go
```diff
@@ -0,0 +1,503 @@
...
+code returned by the service in code. The error code can be used
+to switch on context specific functionality. In this case a context
+specific error message is printed to the user based on the bucket
+and key existing.
+For information on other S3 API error codes see:
+https://aws.github.io/aws-sdk-go-v2/docs/handling-errors/
+*/
+func handleAwsError(err error) error {
+	var oe *smithy.OperationError
+	var ae smithy.APIError
```
cmars, line 482: nit:

```suggestion
	var opErr *smithy.OperationError
	var apiErr smithy.APIError
```

cmars (approved):
More minor questions & comments below, but nothing blocking, this is generally looking good!

vervet-underground/internal/scraper/scraper.go
genslein: For tests or local, port is unique in identifying different services. Since both accept traffic, both are valid config

vervet-underground/internal/storage/s3/client.go
genslein: I suppose, I worry about leaking the connection tho if h

vervet-underground/internal/storage/s3/client.go
genslein: Yes I missed that

vervet-underground/internal/storage/s3/client.go
genslein: Lol

vervet-underground/internal/storage/s3/client.go
genslein: done

PR #177, opened 2022-04-08
feat: backstage catalog commands

cmars:
Add two commands for integrating Vervet with Backstage.

`vervet backstage update-catalog` adds API entities for each compiled
OpenAPI spec to catalog-info.yaml in the project directory.

`vervet backstage preview-catalog` shows what would be written on
update.

Based on #174


jcsackett (approved):

carissadean (approved):
Will we eventually include a catalog-info.yaml as part of the service templates?

cmars:
> Will we eventually include a catalog-info.yaml as part of the service templates?

Yes. I'm actually surprised we don't already do this. We will!

PR #179, opened 2022-04-08
feat: Package application as a nix flake

jgresty:
Nix is cool, flakes are the future, the cool future is now for this
repository. This will allow nix flake users to run the tool by using:

nix run github:snyk/vervet

This should be compatible with plain nix but I haven't tested that.

### DIFF
--- a/default.nix
+++ b/default.nix
@@ -1,10 +1,17 @@
-{ pkgs ? import <nixpkgs> {} }:
-  pkgs.mkShell {
-    nativeBuildInputs = with pkgs.buildPackages; [
-      go envsubst
-    ];
-    shellHook = ''
-      export GOPATH="$HOME/.cache/gopaths/$(sha256sum <<<$(pwd) | awk '{print $1}')"
-    '';
-}
+{ buildGoModule, lib }:
 
+buildGoModule rec {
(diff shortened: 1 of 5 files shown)


default.nix
```diff
@@ -1,10 +1,17 @@
...
-    ];
-    shellHook = ''
-      export GOPATH="$HOME/.cache/gopaths/$(sha256sum <<<$(pwd) | awk '{print $1}')"
-    '';
-}
+{ buildGoModule, lib }:
+
+buildGoModule rec {
+  pname = "vervet";
+  version = "4.6.6";
```
cmars, line 5: I wonder if there's a way to tie this to the latest release tag (can be derived with `ccv`)?

cmars (approved):
LGTM & thanks for this, it's about time I started looking into flakes!

Package build works on darwin if I do:

    export NIXPKGS_ALLOW_UNSUPPORTED_SYSTEM=1
    nix-build -E 'with import <nixpkgs> {}; callPackage ./default.nix {}'

One thing I noticed is the built binary doesn't have the go generated version baked into it. This is also a problem for `go install`ers though.

I'm keen on landing this and testing on pure NixOS in my lab.

PR #181, opened 2022-04-11
feat: storage collator to collate specs and versions

davidharrigan:
This PR adds `storage.Collator` to decouple content revision logic currently implemented inside `mem.Storage` so the same logic can be shared in other storage implementations.

New storage implementations should use `storage.NewCollator()` in their `CollateVersions` to load the collator instance with spec revisions from their storage. `Collator.Collate` outputs unified API versions and specs that the storage implementation can save and use.


vervet-underground/internal/storage/mem/mem.go
```diff
@@ -134,11 +122,11 @@ func (s *Storage) NotifyVersion(name string, version string, contents []byte, sc
 	// End of initializations
 
 	// add the new ContentRevision
-	s.serviceVersionMappedRevisionSpecs[name][version][digest] = ContentRevision{
-		serviceVersion: name + "_" + version,
```
davidharrigan, line 138: This format did not seem to be used except in error logs - so I have removed it, but can include the `service` name here as well if we want to keep it.

vervet-underground/internal/storage/mem/mem.go
```diff
@@ -134,11 +122,12 @@ func (s *Storage) NotifyVersion(name string, version string, contents []byte, sc
...
-		timestamp:      scrapeTime,
-		digest:         digest,
-		blob:           contents,
+	s.serviceVersionMappedRevisionSpecs[name][version][digest] = storage.ContentRevision{
```
genslein, line 125: This may be a subsequent Abstraction for "Storage" interface for when cloud storage mechanisms are ready. It took me a minute to remember how this worked. Can you add a TODO comment here?

vervet-underground/internal/storage/revision.go
```diff
@@ -0,0 +1,69 @@
...
+		Versions:  make(vervet.VersionSlice, 0),
+	}
+}
+
+// Add registers a new ContentRevision for the service.
+func (s *ServiceRevisions) Add(revision ContentRevision) {
+	version := revision.Version
+	if _, ok := s.Revisions[version]; !ok {
+		s.Versions = append(s.Versions, version)
+		sort.Sort(s.Versions)
```
genslein, line 41: Revisions are done by timestamp, they are not post-fixed with stabilities due to scraping intervals. May need to revert in favor of `timestamps`

genslein: ignore, not the intended use, this is for a single service's versionslice support

vervet-underground/internal/storage/collator.go
```diff
@@ -0,0 +1,89 @@
...
+
+	"github.com/snyk/vervet"
+)
+
+// Collator is an aggregate of service specs and versions scraped by VU. It is responsible for collating versions and
+// specs from all services VU manages.
+// This is the top level resource all storage classes should use for producing collated data.
+type Collator struct {
+	revisions map[string]*ServiceRevisions
+	versions  map[vervet.Version]struct{}
```
genslein, line 18: Let's move this to a `VersionSlice` since the struct isn't needed and the sort happens in `Collate`

genslein (approved):

vervet-underground/internal/storage/revision.go
```diff
@@ -0,0 +1,77 @@
...
+	Revisions map[vervet.Version][]ContentRevision
+	// Versions is a collection of API uniqueVersions that this service serves.
+	Versions  vervet.VersionSlice
+}
+
+// NewServiceRevisions returns a new instance of ServiceRevisions
+func NewServiceRevisions() *ServiceRevisions {
+	return &ServiceRevisions{
+		Revisions: make(map[vervet.Version][]ContentRevision),
+		Versions:  make(vervet.VersionSlice, 0),
```
cmars, line 39: The empty value (nil) for a slice should generally be well-tolerated as a zero-length slice, this should not be strictly necessary, though harmless.

vervet-underground/internal/storage/collator.go
```diff
@@ -0,0 +1,101 @@
...
+		src, err := loader.LoadFromData(revision.Blob)
+		if err != nil {
+			return nil, fmt.Errorf("could not merge revision %s-%s-%s: %w", revision.Service, revision.Version, revision.Digest, err)
+		}
+
+		if dst == nil {
+			dst = src
+		} else {
+			// TODO: evaluate whether to use replace bool or not during merging
+			vervet.Merge(dst, src, true)
```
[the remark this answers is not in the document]
cmars: See https://github.com/snyk/vervet/tree/main/testdata/conflict-components for an example of component conflict test cases (used in ./internal/cmd/compiler_test.go)

vervet-underground/internal/storage/collator.go
```diff
@@ -0,0 +1,101 @@
...
+	specs := make(map[vervet.Version]openapi3.T)
+	sort.Sort(c.uniqueVersions)
+
+	for _, version := range c.uniqueVersions {
+		revisions := make([]ContentRevision, 0)
+		for service, serviceRevisions := range c.revisions {
+			rev, err := serviceRevisions.ResolveLatestRevision(version)
+			if err != nil {
+				log.Warn().Err(err).Msgf("could not resolve version %s for service %s", version, service)
+				// don't halt execution if we can't resolve version for this service.
```
cmars, line 64: Why is this recoverable?

vervet-underground/internal/storage/collator.go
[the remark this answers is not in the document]
cmars: See also https://pkg.go.dev/github.com/snyk/vervet/v4#Collator

vervet-underground/internal/storage/collator.go
davidharrigan: It's possible that a version doesn't exist for a service. If service A has version `2022-04-12` and service B with `2022-04-13` - `2022-04-13` would resolve for both services, but `2022-04-12` only for service A.  So if resolving fails here and continue with others. I can make this a bit more clear here.

vervet-underground/internal/storage/collator.go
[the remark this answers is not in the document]
davidharrigan: awesome, thanks for this! It does sound like something we'll need to follow up on. At least adding an alert to start with so we're notified when VU complains here.

vervet-underground/internal/storage/revision.go
```diff
@@ -0,0 +1,77 @@
...
+)
+
+// ContentRevision is the exact contents and metadata of a service's version at scraping timestamp.
+type ContentRevision struct {
+	// Service is the name of the service.
+	Service   string
+	// Version is the API version number of this content.
+	Version   vervet.Version
+	// Timestamp represents when this revision was scraped.
+	Timestamp time.Time
```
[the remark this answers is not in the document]
davidharrigan: Hmm - I'm not sure if we'd ever modify these. If there is a new revision, it will be added as a new content and not overwrite the previous revision. We may find that we'll want modified time as well when we add non-memory storage - so I'll defer this until then.

vervet-underground/internal/storage/collator.go
```diff
@@ -0,0 +1,101 @@
...
+	// Track service and its revision
+	if _, ok := c.revisions[service]; !ok {
+		c.revisions[service] = NewServiceRevisions()
+	}
+	c.revisions[service].Add(revision)
+
+	// Track versions
+	version := revision.Version
+	var found bool
+	for _, v := range c.uniqueVersions{
```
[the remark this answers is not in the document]
davidharrigan: It does not look like it - I can take a look as a follow up.

vervet-underground/internal/storage/collator.go
```diff
@@ -0,0 +1,102 @@
...
+func (c Collator) Collate() (vervet.VersionSlice, map[vervet.Version]openapi3.T, error) {
+	specs := make(map[vervet.Version]openapi3.T)
+	sort.Sort(c.uniqueVersions)
+
+	for _, version := range c.uniqueVersions {
+		revisions := make([]ContentRevision, 0)
+		for service, serviceRevisions := range c.revisions {
+			rev, err := serviceRevisions.ResolveLatestRevision(version)
+			if err != nil {
+				// don't halt execution if we can't resolve version for this service - it is possible for a service to not have this version available.
```
cmars, line 62: Is this because the service might not have had any releases at the time of `version`? That makes sense... I wonder if it should even be a warning if that's the case though -- maybe a debug-level message?

vervet-underground/internal/storage/collator.go
```diff
@@ -0,0 +1,102 @@
...
+				log.Warn().Err(err).Msgf("could not resolve version %s for service %s", version, service)
+				continue
+			}
+			revisions = append(revisions, rev)
+		}
+
+		if len(revisions) > 0 {
+			spec, err := mergeRevisions(revisions)
```
cmars, line 70: A test case with a colliding component would be great!

cmars (approved):
Looks good!

vervet-underground/internal/storage/collator.go
davidharrigan: yep exactly... https://github.com/snyk/vervet/pull/181#discussion_r848734996
You are definitely right, `warn` does seem aggressive here. I think debug will do!

PR #182, opened 2022-04-12
fix: Backstage API entity compatibility

cmars:
A few fixes needed in order to generate API entities that Backstage will
accept:
- API metadata name must match ([A-Z][a-z][0-9]_-)+
- Annotations cannot nest, only one level supported.

Added a more readable, user-friendly title.

Version information is placed under labels.

Tags are also added, derived from versioning fields.

### DIFF
--- a/internal/backstage/backstage.go
+++ b/internal/backstage/backstage.go
@@ -9,6 +9,8 @@ import (
 	"io/fs"
 	"os"
 	"path/filepath"
+	"sort"
+	"strings"
 	"time"
 
 	"github.com/vmware-labs/yaml-jsonpath/pkg/yamlpath"
@@ -19,11 +21,11 @@ import (
 
 const (
 	backstageVersion   = "backstage.io/v1alpha1"
-	snykApiVersion     = "snyk.io/vervet/version"
-	snykApiVersionDate = "snyk.io/vervet/version/date"
-	snykApiStability   = "snyk.io/vervet/version/stability"
-	snykApiLifecycle   = "snyk.io/vervet/version/lifecycle"
-	snykApiGeneratedBy = "snyk.io/vervet/generated-by"
+	snykApiVersion     = "api.snyk.io/version"
+	snykApiVersionDate = "api.snyk.io/version-date"
+	snykApiStability   = "api.snyk.io/version-stability"
+	snykApiLifecycle   = "api.snyk.io/version-lifecycle"
+	snykApiGeneratedBy = "api.snyk.io/generated-by"
 )
 
 // Component represents a Backstage Component entity document.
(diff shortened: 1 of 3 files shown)


internal/backstage/backstage.go
```diff
@@ -265,6 +290,21 @@ func (c *CatalogInfo) vervetAPI(doc *vervet.Document, root string) (*API, error)
...
+		if r >= '0' && r <= '9' {
+			return r
+		}
+		if r >= 'A' && r <= 'Z' {
+			return r
+		}
+		if r >= 'a' && r <= 'z' {
+			return r
+		}
+		return '-'
```
jcsackett, line 304: Do really want to replace everything else with `-`? Maybe just drop them? Or replace them with `-` if they have an alphanum to either side?

I suppose the latter approach would require regex processing which would be Bad(TM). But maybe just replacing spaces and ~ with the -, and drop other characters? The example in the test with a ton of trailing dashes just looks weird to me, but I'm willing to be told I'm overthinking this and that sort of case isn't likely to actually come up.

internal/backstage/backstage_test.go
```diff
@@ -102,13 +121,13 @@ func TestLoadVersionsNoApis(t *testing.T) {
...
-    - Registry 2021-06-13~beta
-    - Registry 2021-06-13~experimental
-    - Registry 2021-08-20~beta
-    - Registry 2021-08-20~experimental
+    - Registry_2021-06-01_experimental
+    - Registry_2021-06-04_experimental
+    - Registry_2021-06-07_experimental
+    - Registry_2021-06-13_beta
+    - Registry_2021-06-13_experimental
+    - Registry_2021-08-20_beta
```
jcsackett, line 129: Shouldn't these be e.g. `Registry-2021-08-20-beta`? Aren't underscores a replaced character? Or am I now mixing up which parts are changed?

jcsackett (approved):
LGTM, though I have some questions in the diff.

internal/backstage/backstage.go
cmars: I agree -- one of my unit tests already kind of makes this point as well, thanks for raising this!

internal/backstage/backstage_test.go
cmars: I decided to replace invalid characters in the API name prefix (derived from .info.title in the OpenAPI spec) with dashes, then join that with the version date and stability using underscores. Following your above suggestion, I think I might replace only spaces with dashes, and drop all other invalid chars.

This has a nice property of being able to split the name on '_' into three parts (name, version date, stability). If I use dashes to join the name and version, parsing is still possible, but gets a little more complicated.

I'm not quite sure what I'm going to be able to pull out of Backstage in various integrations to do reporting, so I'm leaving room for some optionality here.

jcsackett: Ah, ok, so this is the final product of the join--I had missed that detail. And yes, I can see the utility of having a uniform separator that also scans pretty well in titles.

PR #184, opened 2022-04-13
chore: use viper to drive vu config

davidharrigan:
This PR updates VU config structure to use viper to drive configuration and default values. 

[GRFE-512](https://snyksec.atlassian.net/browse/GRFE-512)


vervet-underground/config/config.go
```diff
@@ -2,26 +2,42 @@
...
 // config file.
-func Load(configPath string) (*ServerConfig, error) {
-	file, err := os.Open(configPath)
-	var config ServerConfig
-	if err != nil {
+func Load(configPaths ...string) (*ServerConfig, error) {
+	setDefaults()
+
+	for _, c := range configPaths {
+		viper.SetConfigFile(c)
```
jlourenc, line 33: Does `viper` accept multiple config files? Or did you mean to call `AddConfigPath` ?

vervet-underground/config/config.go
```diff
@@ -2,26 +2,42 @@
...
-	Host     string   `json:"host"`
-	Services []string `json:"services"`
+	Host           string
+	Services       []string
+	StorageBackend StorageBackendType
 }
 
-// Load returns a ServerConfig instance loaded from the given path to a JSON
+func setDefaults() {
+	viper.SetDefault("host", "0.0.0.0")
```
jlourenc, line 23: I'm a bit hesitant to make this the default as it means listening on a specific port on all interfaces for a server. Would `localhost` be a good enough alternative?

vervet-underground/config/config_test.go
```diff
@@ -0,0 +1,89 @@
...
+			Host:           "0.0.0.0",
+			Services:       []string{"localhost"},
+			StorageBackend: config.StorageBackendMemory,
+		}
+		c.Assert(*conf, qt.DeepEquals, expected)
+	})
+
+	c.Run("multiple configs", func(c *qt.C) {
+		defaultConfig, err := createTestFile([]byte(`{
+			"host": "0.0.0.0",
```
jlourenc, line 66: I think setting `host` to `localhost` would not give you `localhost` as a result once the config parsed.

vervet-underground/config/config_test.go
```diff
@@ -0,0 +1,89 @@
...
+	"os"
+	"testing"
+
+	qt "github.com/frankban/quicktest"
+
+	"vervet-underground/config"
+)
+
+func createTestFile(data []byte) (*os.File, error) {
+	f, err := os.CreateTemp("", "tmpfile-*.json")
```
genslein, line 13: Thank you for using json 👍 😄 😭

genslein (approved):
Once @jlourenc comments are addressed

vervet-underground/config/config.go
davidharrigan: yes it does (there is a test for it too). `AddConfigPath` just tells viper which path to find config files from, but not which config file.

vervet-underground/config/config_test.go
davidharrigan: yes you're right - looks like I can use `MergeInConfig` to avoid that issue. I'll fix it and update the test. Thank you for the spot!

vervet-underground/go.mod
```diff
@@ -31,25 +32,47 @@ require (
...
 	golang.org/x/sys v0.0.0-20220114195835-da31bd327af9 // indirect
 	golang.org/x/text v0.3.7 // indirect
 	golang.org/x/xerrors v0.0.0-20220411194840-2f41105eb62f // indirect
 	gopkg.in/check.v1 v1.0.0-20201130134442-10cb98267c6c // indirect
+	gopkg.in/ini.v1 v1.66.2 // indirect
 	gopkg.in/tomb.v1 v1.0.0-20141024135613-dd632973f1e7 // indirect
 	gopkg.in/yaml.v2 v2.4.0 // indirect
 )
+
+replace (
```
cmars, line 68: If you update to the latest viper (v1.11.0) are these still needed? Was brushing up on viper as it'd been awhile and noticed they just released yesterday.

cmars (approved):
LGTM, a few thoughts below.

cmars:
Also wondering if a `go mod tidy` might trim the go.sum a bit?

vervet-underground/go.mod
davidharrigan: oh neat! that might help us here, I will take a look.

davidharrigan:
Unfortunately that is as tidy as it gets for go.sum :/

genslein (approved):

davidharrigan:
License check issue is ok-ed internally. I will merge this PR with the failing checks, but will suppress them from the project settings.

PR #186, opened 2022-04-14
chore: collect metrics from scraper

davidharrigan:
This PR adds prometheus metrics for Vervet Underground.

- `/metrics` endpoint publishes prom metrics.
- scraper run duration and errors are recorded.
- scrape duration and errors per service are measured.
- outbound requests made from scraper are measured.
- collator errors are measured.
- request/response duration for the VU http server are measured.


vervet-underground/internal/scraper/metrics.go
```diff
@@ -0,0 +1,70 @@
...
+		Help:    "Time spent on a service http call",
+		Buckets: []float64{0.025, 0.05, 0.1, 0.25, 0.5, 1, 2.5, 5, 10},
+	}, []string{"host", "endpoint", "method", "status"})
+)
+
+// durationAllowList is a list of regex matchers for patterns VU collect metrics for outbound requests.
+var durationAllowList = []*regexp.Regexp{
+	regexp.MustCompile(`/openapi$`),
+	regexp.MustCompile(`/openapi/[1-9][0-9]{3}-[0-1][0-9]-[0-3][0-9](~\w+)?$`),
+}
```
davidharrigan, line 47: even with the allow list - I'm wondering if this could be too much? 
if VU scrapes 20 services, that could be 20 * 365 versions * 4 stability levels. Maybe removing the endpoint label would be preferred here - I would expect all endpoints to perform similarly as they should just be returning static open API specs.

vervet-underground/metrics.go
```diff
@@ -0,0 +1,47 @@
...
+type responseWriter struct {
+	http.ResponseWriter
+	statusCode int
+}
+
+// WriteHeader writes status code to the http.ResponseWriter, as well as intercepting the status code for use within
+// the middleware.
+func (w *responseWriter) WriteHeader(statusCode int) {
+	w.statusCode = statusCode
+	w.ResponseWriter.WriteHeader(statusCode)
```
jlourenc, line 29:
```suggestion
	w.ResponseWriter.WriteHeader(statusCode)
	if w.statusCode == 0 {
		w.StatusCode = statusCode
	}
```
The reason for inverting the calls is that `w.ResponseWriter.WriteHeader` panics is the. status code is invalid which can be recovered and handled again by the business logic.

The reason for having a conditional is that `w.ResponseWriter.WriterHeader` only actions the first valid call to it so a status code can only be set once.

vervet-underground/internal/scraper/scraper.go
```diff
@@ -140,6 +153,11 @@ func (s *Scraper) Run(ctx context.Context) error {
 		err := s.collateVersions()
 		errs = multierr.Append(errs, err)
 	}
+
+	scraperRunDuration.Observe(time.Since(scrapeTime).Seconds())
```
jlourenc, line 157: This won't be done if there are scraping errors. I think we need to refactor the error handling just above.

vervet-underground/internal/scraper/scraper.go
```diff
@@ -140,6 +153,11 @@ func (s *Scraper) Run(ctx context.Context) error {
...
+	if errs != nil {
+		scraperRunError.Inc()
```
jlourenc, line 159: I think this is only incrementing when there are `collateVersions` errors since there is an early return line 150.

vervet-underground/internal/scraper/metrics.go
```diff
@@ -0,0 +1,70 @@
...
+	})
+
+	// Scraper service scrape metrics.
+	serviceScrapeDuration = promauto.NewHistogramVec(prometheus.HistogramOpts{
+		Name:    "vu_scraper_service_scrape_duration_seconds",
+		Help:    "Time spent scraping a service",
+		Buckets: []float64{1, 5, 10, 25, 45, 60, 90},
+	}, []string{"service"})
+	serviceScrapeError = promauto.NewCounterVec(prometheus.CounterOpts{
+		Name: "vu_scraper_error_total",
```
jlourenc, line 33: nit: Should this be named `vu_scraper_service_scrape_error_total` to align with `vu_scraper_service_scrape_duration_seconds` ?

vervet-underground/internal/scraper/metrics.go
```diff
@@ -0,0 +1,70 @@
...
+}
+
+// DurationTransport returns a http.RoundTripper for tracking http calls made from the Scraper.
+// Expects a histogram vector with host, endpoint, method, and status labels.
+func DurationTransport(next http.RoundTripper) http.RoundTripper {
+	return promhttp.RoundTripperFunc(func(r *http.Request) (*http.Response, error) {
+		start := time.Now()
+		resp, err := next.RoundTrip(r)
+		if err == nil && pathAllowed(durationAllowList, r.URL.Path) {
+			status := fmt.Sprintf("%dxx", resp.StatusCode/100)
```
jlourenc, line 56: 🧠  nice

vervet-underground/internal/scraper/metrics.go
```diff
@@ -0,0 +1,70 @@
...
+		Name:    "vu_scraper_http_duration_seconds",
+		Help:    "Time spent on a service http call",
+		Buckets: []float64{0.025, 0.05, 0.1, 0.25, 0.5, 1, 2.5, 5, 10},
+	}, []string{"host", "endpoint", "method", "status"})
+)
+
+// durationAllowList is a list of regex matchers for patterns VU collect metrics for outbound requests.
+var durationAllowList = []*regexp.Regexp{
+	regexp.MustCompile(`/openapi$`),
+	regexp.MustCompile(`/openapi/[1-9][0-9]{3}-[0-1][0-9]-[0-3][0-9](~\w+)?$`),
```
genslein, line 46: `w+` can be an explicit `(wip|experimental|beta)` arbitrary words are probably not so great for cardinality.

vervet-underground/metrics.go
```diff
@@ -0,0 +1,47 @@
...
+	"strconv"
+	"time"
+
+	"github.com/gorilla/mux"
+	"github.com/prometheus/client_golang/prometheus"
+	"github.com/prometheus/client_golang/prometheus/promauto"
+)
+
+// VU server metrics.
+var httpDuration = promauto.NewHistogramVec(prometheus.HistogramOpts{
```
genslein, line 14: are we not using https://github.com/slok/go-http-metrics? This should probably be a standard emission.

vervet-underground/metrics.go
```diff
@@ -0,0 +1,47 @@
...
+// WriteHeader writes status code to the http.ResponseWriter, as well as intercepting the status code for use within
+// the middleware.
+func (w *responseWriter) WriteHeader(statusCode int) {
+	w.statusCode = statusCode
+	w.ResponseWriter.WriteHeader(statusCode)
+}
+
+// prometheusMiddleware tracks duration of the wrapped http.Handler. Metric is only collected if the path template has
+// been registered with mux.
+func prometheusMiddleware(next http.Handler) http.Handler {
```
genslein, line 34: We do use https://github.com/slok/go-http-metrics in other applications I think we can use the standardization here.

vervet-underground/internal/scraper/scraper.go
davidharrigan: ah yeah good catch. I missed the return statement in L151.

vervet-underground/internal/scraper/metrics.go
cmars: What if instead of a regex, trim off the `.*/openapi` prefix, and use `vervet.ParseVersion` to check the rest? That way if/when we define new stability levels (`lts`?) all that's needed is a dependency update here.

vervet-underground/internal/scraper/metrics.go
cmars: What do you care about tracking here? If this is the duration of time it takes to scrape an OpenAPI spec, it might be reasonable to just tag the service, and rely on logs to drill into why / where a request is slow.

Generally services should be serving these specs from memory -- in Go we use embed.FS, in Node we pre-load them all on init for a a middleware closure.

cmars: So maybe we don't even need the regex -- we can just lump all the /openapi paths together possibly?

cmars (commented):
Some thoughts below on the metric labels...

vervet-underground/internal/scraper/metrics.go
davidharrigan: Yeah - just to track if there's any slowness in fetching specs from a particular service - I agree that lumping /openapi paths together and relying on logs to dive in further is probably ok here. I'll make that change!

vervet-underground/internal/scraper/metrics.go
genslein: I think that's preferred

vervet-underground/metrics.go
davidharrigan: this is now removed

vervet-underground/internal/scraper/metrics.go
davidharrigan: I've left the regex as is for now - we're no longer including the `endpoint` label based on this discussion: https://github.com/snyk/vervet/pull/186#discussion_r850491186

genslein (approved):
@davidharrigan  can we get a follow-up PR to add a prom + grafana to `docker-compose.yml` setup so we can validate metrics locally?

vervet-underground/internal/scraper/metrics.go
cmars: This is fair -- we should probably be validating the response of `/openapi` before making the requests anyway (are we?).

cmars (approved):

PR #187, opened 2022-04-15
feat: use fs.FS for generators

jcsackett:
rest-go-generate uses vervet's generator code to produce code for go
projects. To create a single binary, the generators.yaml and template
files for rest-go-generate are provided with embed.FS

To avoid special casing, vervet now always loads generation files from
an fs.FS object, using os.DirFS mounted at root if no other option is
provided.

### DIFF
--- a/internal/generator/generator.go
+++ b/internal/generator/generator.go
@@ -153,7 +153,7 @@ func (g *Generator) resolveFilename(filenameTemplate string) (string, error) {
 	if err != nil {
 		return "", err
 	}
-	return buf.String(), nil
+	return filepath.Abs(buf.String())
 }
 
 // Option configures a Generator.
--- a/internal/generator/generator.go
+++ b/internal/generator/generator.go
@@ -90,7 +90,7 @@ func New(conf *config.Generator, options ...Option) (*Generator, error) {
 		}
 	}
 
-	// Resolve the template filename... with a template.  Only .Here is and .Cwd
+	// Resolve the template filename... with a template.  Only .Here and .Cwd
 	// are supported, not full scope. Just enough to locate files relative to
 	// the config.
 	templateFilename, err := g.resolveFilename(conf.Template)
--- a/internal/generator/generator.go
+++ b/internal/generator/generator.go
@@ -4,10 +4,12 @@ import (
 	"bytes"
 	"fmt"
 	"io"
+	"io/fs"
 	"io/ioutil"
 	"log"
 	"os"
 	"path/filepath"
+	"strings"
 	"text/template"
 
 	"github.com/ghodss/yaml"
@@ -29,6 +31,7 @@ type Generator struct {
 	dryRun bool
 	force  bool
 	here   string
+	fs     fs.FS
 }
 
 // NewMap instanstiates a map of Generators from configuration.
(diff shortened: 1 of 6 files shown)


internal/generator/generator.go
```diff
@@ -81,17 +90,24 @@ func New(conf *config.Generator, options ...Option) (*Generator, error) {
 		}
 	}
 
-	// Resolve the template filename... with a template. Only .Here is
-	// supported, not full scope. Just enough to locate files relative to the
-	// config.
+	// Resolve the template filename... with a template.  Only .Here is and .Cwd
```
cmars, line 93:
```suggestion
	// Resolve the template filename... with a template.  Only .Here and .Cwd
```

internal/generator/generator_test.go
```diff
@@ -40,7 +40,7 @@ func TestVersionScope(t *testing.T) {
 version-readme:
   scope: version
   filename: "{{.Here}}/{{.API}}/{{.Resource}}/{{.Version.DateString}}/README"
-  template: ".vervet/resource/version/README.tmpl"
+  template: "{{.Cwd}}/.vervet/resource/version/README.tmpl"
```
cmars, line 43: Without this change, if the template path didn't have the `{{.Cwd}}/` prefix, will this still pass if `os.DirFS` is used?

If so, can we have this case tested as well?

cmars (commented):
Almost LGTM, just a question below.

internal/generator/generator_test.go
jcsackett: It won't; that would be relative path, which doesn't work with `os.DirFS`--we could remove the Cwd thing, and check for a "/" prefix instead, and if there's not one do the change to the filepath change there *if* we're using DirFS, but then we're special casing for DirFS.

jcsackett: Instead of Cwd we could provide `{{ProjectDir}}` and have folks use that for paths relative to the projectdir (which is *usually* the wd as well). That actually sounds a bit better, since it's defined as the directory the vervet config is in, rather than changing depending where you're running the command.

It would require passing around some more params though, and we're getting to some beefy structs for those.

cmars: I'm OK with adding `{{.Cwd}}` as it's implemented here, but I think a special case to effectively prefix a relative path with the cwd when DirFS is used, would be reasonable. Otherwise we're breaking the current behavior of how generator templates can be currently referenced in 4.x.

If it's terribly difficult I might be convinced to look the other way and make the breaking change. Vervet v4 and this functionality is released. While I don't think this breaking change would affect rest-node-generate (since it references all the templates off of `{{.Here}}` in a node project), I'd feel slightly bad about it.

cmars (approved):
LGTM. I really like making the template paths explicit with `{{.Cwd}}`. Noting below that there might be a technically breaking change here, but I don't think it is being used anywhere.

My only ask is that you confirm this works with rest-node-generate before landing. Thanks for this!

internal/generator/generator_test.go
jcsackett: I concur about breaking changes; rest-node-generate is using abs paths off of .Here, but we don't know what other folks are doing. 

We can leave a note to remove the special casing whenever we release v5.

cmars: Turns out supporting both is fairly straightforward; just needed to `filepath.Abs` after interpolating the filename (3a9d7462f3018ab8c6560fa4016639b34c315195)

cmars:
Confirmed this works with rest-node-generate. Given no regressions there and the non-os.DirFS unit test added, I think this is good to land now.

PR #188, opened 2022-04-18
chore: testing collator validation missing actual collation

genslein:
We are not actually collating specs from services. Need to debug

image

https://github.com/snyk/vervet/blob/main/vervet-underground/internal/storage/collator.go#L63

We're failing every time

### DIFF
--- a/vervet-underground/internal/scraper/scraper.go
+++ b/vervet-underground/internal/scraper/scraper.go
@@ -148,10 +148,7 @@ func (s *Scraper) scrape(ctx context.Context, scrapeTime time.Time, svc service)
 	if err != nil {
 		return errors.WithStack(err)
 	}
-	//err = s.storage.NotifyVersions(svc.base, versions, scrapeTime)
-	//if err != nil {
-	//	return errors.WithStack(err)
-	//}
+
 	for i := range versions {
 		// TODO: we might run this concurrently per live service pod if/when
 		// we're more k8s aware, but we won't do that yet.
--- a/vervet-underground/internal/scraper/scraper.go
+++ b/vervet-underground/internal/scraper/scraper.go
@@ -87,7 +87,7 @@ func setupScraper(s *Scraper, cfg *config.ServerConfig, options []Option) error 
 		if err != nil {
 			return errors.Wrapf(err, "invalid service %q", cfg.Services[i])
 		}
-		s.services[i] = service{base: cfg.Services[i], url: u}
+		s.services[i] = service{base: u.Host, url: u}
 	}
 	for i := range options {
 		err := options[i](s)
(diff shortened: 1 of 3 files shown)


vervet-underground/internal/scraper/scraper_test.go
```diff
@@ -104,6 +105,28 @@ func TestScraper(t *testing.T) {
...
+		case "2021-10-01":
+			c.Assert(len(spec.Paths), qt.Equals, 3)
+		case "2021-10-16":
+			c.Assert(len(spec.Paths), qt.Equals, 4)
+		case "2021-09-01":
+			c.Assert(len(spec.Paths), qt.Equals, 1)
+		case "2021-09-16":
+			c.Assert(len(spec.Paths), qt.Equals, 2)
+		default:
+			c.FailNow()
```
cmars, line 127: This would be better as a table test.

vervet-underground/internal/scraper/scraper_test.go
```diff
@@ -211,4 +234,26 @@ func TestScraperCollation(t *testing.T) {
 	collated, err := st.GetCollatedVersionSpecs()
 	c.Assert(err, qt.IsNil)
 	c.Assert(len(collated), qt.Equals, 4)
+
+	c.Assert(len(st.Versions()), qt.Equals, 4)
```
cmars, line 238: Same here

vervet-underground/internal/scraper/scraper.go
```diff
@@ -148,10 +148,10 @@ func (s *Scraper) scrape(ctx context.Context, scrapeTime time.Time, svc service)
...
 		return errors.WithStack(err)
 	}
-	err = s.storage.NotifyVersions(svc.base, versions, scrapeTime)
-	if err != nil {
-		return errors.WithStack(err)
-	}
+	//err = s.storage.NotifyVersions(svc.base, versions, scrapeTime)
+	//if err != nil {
+	//	return errors.WithStack(err)
+	//}
```
cmars, line 154: Why is this removed? I hope it's not because we're trying to debug this in preprod... we'd be better off reproducing in a local test that serves up openapi specs with httptest to be scraped...

vervet-underground/internal/scraper/scraper_test.go
cmars: Or a map[string]int of "# expected paths" perhaps?

vervet-underground/internal/scraper/scraper.go
genslein: Because this actually does nothing and in fact, wipes out local mem when done. We don't have any content since `getNewVersion` hasn't been called so this is effectively not helping at all. Line 166 is where the actual work is being done.

cmars: Should we just remove it then?

cmars (approved):

vervet-underground/internal/scraper/scraper.go
cmars: Ok, already was :)

vervet-underground/internal/scraper/scraper_test.go
genslein: done

vervet-underground/internal/scraper/scraper_test.go
genslein: done

PR #189, opened 2022-04-20
Kubernetes service discovery

davidharrigan:
This PR adds the ability for vervet-underground to discover services to scrape inside of a Kubernetes cluster.  This PR extends great work already done by @Todai88 in #114.

To enable:
- set `serviceDiscoveryEnabled` config to `true`.
- add the following annotations for your k8s service resource. VU looks for these annotations to register the service with VU.
  - `vervet-underground.snyk.io/scrape: true`
  - `vervet-underground.snyk.io/port: 8080` (defaults to 80)

Tiltfile to deploy VU and a demo service / make target to spin up a k3d cluster has been added to demonstrate the discovery behavior.

Two new concepts: `service.Registry` and `service.Loader` have been added to the `service` package. `service.Registry` is a store of services that VU should scrape.  The store is refreshed each time scraping begins - so VU attempts to re-discover services on each scraper run.
`service.Loader` describes how to load services into `service.Registry`. Two implementations `StaticServiceLoad` which loads service names from config and `KubeServiceLoader` which discovers services from k8s cluster are added.


vervet-underground/server.go
```diff
@@ -54,19 +68,20 @@ func main() {
...
 	}
 
-	sc, err := scraper.New(cfg, st, scraper.HTTPClient(&http.Client{Timeout: wait}))
+	sc, err := scraper.New(registry, st, scraper.HTTPClient(&http.Client{Timeout: wait}))
 	if err != nil {
 		logError(err)
 		log.Fatal().Msg("unable to load storage")
 	}
 	// initialize
+	// TODO: track service health instead of shutting down the scraper
```
davidharrigan, line 77: I think a further improvement here would be to track whether service scraping succeeded for each service rather than halting vu from scraping. Though that might get complicated if 1 service being offline could remove entries from collated OpenAPI spec.

davidharrigan:
oh no it's the same license check... I think I'm going to cry 😭 . (We're working through this currently for the other PRs)

vervet-underground/internal/service/loader_test.go
```diff
@@ -0,0 +1,35 @@
+package service_test
+
+import (
+	"context"
+	"testing"
+
+	qt "github.com/frankban/quicktest"
+
+	. "vervet-underground/internal/service"
```
genslein, line 9: We'll need to remove the dot import

vervet-underground/internal/service/registry.go
```diff
@@ -0,0 +1,78 @@
...
+	"fmt"
+	"net/url"
+	"sync"
+)
+
+// Service represents a target for vervet-underground to scrape.
+type Service struct {
+	Base string
+	URL  *url.URL
+	// TODO: track healthcheck?
```
genslein, line 15: would it make sense to have the string in addition to the struct, and omit the struct from Marshal parsing, rather than the custom marshal unmarshal?

vervet-underground/server.go
genslein: https://github.com/snyk/vervet/blob/main/vervet-underground/internal/scraper/scraper.go#L99 we allow all scraping to run in parallel in goroutines. What we do with scrapes on retrieval is up to us. What did you have in mind?

genslein (approved):
Very excited to see we have a plan forward here.

vervet-underground/internal/service/registry.go
davidharrigan: Agreed I think that's a better choice here

vervet-underground/server.go
davidharrigan: My thought here was that as we increase more services being scraped by VU, especially with the addition of service discovery here - chance of VU scrape failing increases (maybe bad port being configured for example).  Maybe we'll want to let VU report which service is "unhealthy" and let it continue with collation?

cmars:
This looks very interesting and I'd like to rebase & pick this up once we get VU stable and collating multiple services.

PR #190, opened 2022-04-22
chore: add local prometheus and grafana

davidharrigan:
This PR adds a simple dashboard and prometheus to VU's docker-compose file.  Will require #186 to browse metrics, but can be merged independently. 

Screen Shot 2022-04-21 at 9 58 02 PM

### DIFF
--- a/vervet-underground/config.default.json
+++ b/vervet-underground/config.default.json
@@ -6,6 +6,6 @@
   ],
   "host": "0.0.0.0",
   "storage": {
-    "type": "mem"
+    "type": "memory"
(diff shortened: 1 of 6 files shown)


Todai88 (approved):
LGTM, but I'm not entirely sure why we think it's important to have metrics available in a dockerised environment. Not because they aren't useful, but rather because I'd have imagined we would run VU in a KinD (not necessarily Kind) cluster for local development.

Anywho, :shipit:

migruealt (approved):

PR #191, opened 2022-04-22
fix: Allow nix package to be build on clean systems

jgresty:
proxyVendor was set which assumes that the go dependencies are available
in the building systems nix store, which is rarely the case. This patch
allows nix to download the dependencies when the requested package is
being built.

We also set the version to the last modified date so this flake will get
a new version number without having to update the version number
manually.

### DIFF
--- a/default.nix
+++ b/default.nix
@@ -1,12 +1,10 @@
-{ buildGoModule, lib }:
-
+{ buildGoModule, lib, lastMod }:
 buildGoModule rec {
   pname = "vervet";
-  version = "4.6.6";
+  version = builtins.substring 0 8 lastMod;
   src = ./.;
 
-  vendorSha256 = null;
-  proxyVendor = true;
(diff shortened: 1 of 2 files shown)


default.nix
```diff
@@ -1,12 +1,10 @@
...
+{ buildGoModule, lib, lastMod }:
 buildGoModule rec {
   pname = "vervet";
-  version = "4.6.6";
+  version = builtins.substring 0 8 lastMod;
   src = ./.;
 
-  vendorSha256 = null;
-  proxyVendor = true;
+  vendorSha256 = "sha256-KcxaAOp9+d2DN4qFmmNDH3Di+TmiqhTbLsujh2aKRmk=";
```
cmars, line 7: Is this going to get out of sync with lastMod?

jgresty: this will get out of sync if any of the dependencies change

cmars (approved):

PR #193, opened 2022-04-26
feat: loads the functions file from provided FS

jcsackett:


internal/generator/functions.go
```diff
@@ -91,7 +91,14 @@ var jsConsole = map[string]func(goja.FunctionCall) goja.Value{
 }
 
 func (g *Generator) loadFunctions(filename string) error {
-	src, err := ioutil.ReadFile(filename)
+	// Remove leading slash, if it exists, from filename. fs.FS.Open does not
+	// accept rooted paths.
+	filename = strings.TrimPrefix(filename, "/")
```
cmars, line 96: Should we do this once in `resolveFilename` instead?

cmars (approved):
Was looking for test coverage, but we have os.DirFS coverage here, I think it's good. One question below.

PR #197, opened 2022-05-03
fix: enable resolve on collated versions

genslein:
This was a miss that prevents the /openapi/experimental lookup
on VU for all storage backends. Required for shared routing

https://github.com/snyk/gloo-edge-api-gateway/pull/374

### DIFF
--- a/vervet-underground/internal/storage/gcs/client.go
+++ b/vervet-underground/internal/storage/gcs/client.go
@@ -6,6 +6,7 @@ import (
 	"errors"
 	"fmt"
 	"io"
+	"sort"
 	"strings"
 	"time"
 
@@ -215,11 +216,28 @@ func (s *Storage) Versions() []string {
 
 // Version retrieves a specific Collated Version.
 func (s *Storage) Version(version string) ([]byte, error) {
-	_, err := vervet.ParseVersion(version)
+	parsedVersion, err := vervet.ParseVersion(version)
 	if err != nil {
 		return nil, err
 	}
-	return s.GetCollatedVersionSpec(version)
+	blob, err := s.GetCollatedVersionSpec(version)
+	if err != nil {
+		collatedVersions := vervet.VersionSlice{}
+		for _, s := range s.Versions() {
+			temp, err := vervet.ParseVersion(s)
+			if err != nil {
+				continue
+			}
+			collatedVersions = append(collatedVersions, temp)
+		}
+		sort.Sort(collatedVersions)
+		resolved, err := collatedVersions.Resolve(parsedVersion)
+		if err != nil {
+			return nil, err
+		}
+		return s.GetCollatedVersionSpec(resolved.String())
+	}
+	return blob, nil
 }
 
 // ListCollatedVersions nice wrapper around the GCS ListObjects request.
--- a/vervet-underground/internal/storage/gcs/client.go
+++ b/vervet-underground/internal/storage/gcs/client.go
@@ -6,8 +6,8 @@ import (
 	"errors"
 	"fmt"
 	"io"
-	"sort"
 	"strings"
+	"sync"
 	"time"
 
 	"cloud.google.com/go/storage"
@@ -39,8 +39,10 @@ type Config struct {
 
 // Storage implements storage.Storage.
 type Storage struct {
-	c      *storage.Client
-	config Config
+	mu               sync.RWMutex
+	c                *storage.Client
+	config           Config
+	collatedVersions vervet.VersionSlice
 }
 
 /*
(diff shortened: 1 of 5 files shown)


vervet-underground/internal/storage/s3/client.go
```diff
@@ -187,11 +188,28 @@ func (s *Storage) Versions() []string {
...
 // Version implements scraper.Storage.
 func (s *Storage) Version(version string) ([]byte, error) {
-	_, err := vervet.ParseVersion(version)
+	parsedVersion, err := vervet.ParseVersion(version)
 	if err != nil {
 		return nil, err
 	}
-	return s.GetCollatedVersionSpec(version)
+	blob, err := s.GetCollatedVersionSpec(version)
+	if err != nil {
```
davidharrigan, line 196: So this first attempts to get a collated version spec for an exact match. If that's not found, then it will try to resolve a version from all available collated versions - specifically for versions like `beta` and `experimental` (didn't realize those were valid options - but makes sense).

This also has a side effect of always attempting to resolve to the closest valid version other than just `beta` and `experimental` - maybe that's ok? 

Maybe this could be pulled out into a function that does parsing / resolving that can be used in all the storage implementations?

vervet-underground/internal/storage/gcs/client.go
```diff
@@ -215,11 +216,28 @@ func (s *Storage) Versions() []string {
...
 		return nil, err
 	}
-	return s.GetCollatedVersionSpec(version)
+	blob, err := s.GetCollatedVersionSpec(version)
+	if err != nil {
+		collatedVersions := vervet.VersionSlice{}
+		for _, s := range s.Versions() {
+			temp, err := vervet.ParseVersion(s)
+			if err != nil {
+				continue
```
cmars, line 229: When/why would the version be an invalid version string?

vervet-underground/internal/storage/s3/client.go
genslein: That's actually more than ok, we should have always been doing that. It's what does the routing algorithm in cerberus as well. This was supposed to be default behavior.

cmars (commented):
I don't understand why versions would fail to parse in this PR.

vervet-underground/internal/storage/gcs/client.go
genslein: It shouldn't necessarily ever, but since we're not holding a collection of Versions for the S3 and GCS implementations cached in memory, this is more respecting the cloud clients responses.  

This is more of a best effort here and I really don't expect to ever run into this case as being a problem. That being said, I'm happy to make an adjustment here from the responses in the cloud clients.

davidharrigan: I wonder if we should return an error here and treat it as a 500 response

davidharrigan (approved):

vervet-underground/internal/storage/gcs/client.go
cmars: I think errors should always be handled in such cases. Log and backoff retry until all successful.

genslein:
```go
		collatedVersions := vervet.VersionSlice{}
		retry := 3
		var versionSlice []string
		for i := 0; i < retry; i++ {
			versionSlice = s.Versions()
			if versionSlice != nil {
				break
			}
		}
		for _, s := range versionSlice {
			for i := 0; i < retry; i++ {
				temp, err := vervet.ParseVersion(s)
				if err != nil {
					log.Debug().Err(err).Msgf("ParseVersion attempt number %d", i)
					continue
				}
				collatedVersions = append(collatedVersions, temp)
				break
			}
		}
```
 The alternate if we fail the `Resolve` in the mem storage is to fail immediately if fuzzy matching breaks. I think I'd rather do that here.

genslein: Alternatively, I can implement the `collatedVersions vervet.VersionSlice` cache in the GCS and S3 clients to handle this.

genslein: I've moved to reusing the `vervet.VersionSlice` caching for endpoint resolution

genslein:
> I don't understand why versions would fail to parse in this PR.

I've removed the client lookups in favor of the local version caching in memory for every client for list of available versions.

cmars (approved):
LGTM.

Most of my comments below are drive-bys concerning context propagation and other things. As such, they are more food for thought / fine for followup than anything blocking.

vervet-underground/internal/storage/gcs/client.go
```diff
@@ -381,16 +407,18 @@ func (s *Storage) ListObjects(key string, delimeter string) ([]storage.ObjectAtt
...
 	bucket, err := s.getBucketAttrs()
-	if err != nil || bucket.Name != s.config.BucketName {
-		if !errors.Is(err, storage.ErrBucketNotExist) {
-			return err
-		}
+	if err != nil && !errors.Is(err, storage.ErrBucketNotExist) {
+		return err
+	}
+	// if bucket exists, idempotent return
+	if bucket != nil {
```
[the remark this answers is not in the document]
genslein: I think this is ok with k8s deployments restarting pods

Issue #198, opened 2022-05-05
Add context propagation to Vervet-Underground 


genslein:
Looks like s3 and GCP clients would also benefit from context propagation
#197 detected

PR #200, opened 2022-05-05
fix: add context propagation to VU

genslein:
https://github.com/snyk/vervet/issues/198

Fixes some stylings from #197 as well

### DIFF
--- a/vervet-underground/internal/storage/gcs/client.go
+++ b/vervet-underground/internal/storage/gcs/client.go
@@ -320,9 +320,6 @@ func (s *Storage) GetCollatedVersionSpec(ctx context.Context, version string) ([
 // PutObject nice wrapper around the GCS PutObject request.
 func (s *Storage) PutObject(ctx context.Context, key string, reader io.Reader) (*storage.ObjectHandle, error) {
 	obj := s.c.Bucket(s.config.BucketName).Object(key)
-	ctx, cancel := context.WithTimeout(ctx, time.Second*50)
-	defer cancel()
-
 	wc := obj.NewWriter(ctx)
 	defer wc.Close()
 
--- a/vervet-underground/internal/storage/gcs/client.go
+++ b/vervet-underground/internal/storage/gcs/client.go
@@ -34,12 +34,14 @@ type Config struct {
 	GcsEndpoint    string
 	BucketName     string
 	IamRoleEnabled bool
+	Context        context.Context
 	Credentials    StaticKeyCredentials
 }
 
 // Storage implements storage.Storage.
 type Storage struct {
 	mu               sync.RWMutex
+	ctx              context.Context
 	c                *storage.Client
 	config           Config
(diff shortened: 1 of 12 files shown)


vervet-underground/internal/storage/gcs/client.go
```diff
@@ -34,12 +34,14 @@ type Config struct {
...
 	BucketName     string
 	IamRoleEnabled bool
+	Context        context.Context
 	Credentials    StaticKeyCredentials
 }
 
 // Storage implements storage.Storage.
 type Storage struct {
 	mu               sync.RWMutex
+	ctx              context.Context
```
cmars, line 44: Context should almost never be stored in a struct. It should be passed to every function or method that propagates it, as the first argument. This means we'll need to change the Storage interface and all its implementations. Some like mem can ignore the context. I think it should be fine to do so, since these are internal packages.

> Programs that use Contexts should follow these rules to keep interfaces consistent across packages and enable static analysis tools to check context propagation:
>
> Do not store Contexts inside a struct type; instead, pass a Context explicitly to each function that needs it. The Context should be the first parameter, typically named ctx: 

(Source: https://pkg.go.dev/context#pkg-overview)

vervet-underground/internal/storage/gcs/client.go
```diff
@@ -319,10 +322,8 @@ func (s *Storage) GetCollatedVersionSpec(version string) ([]byte, error) {
 
 // PutObject nice wrapper around the GCS PutObject request.
 func (s *Storage) PutObject(key string, reader io.Reader) (*storage.ObjectHandle, error) {
-	ctx := context.Background()
 	obj := s.c.Bucket(s.config.BucketName).Object(key)
-
-	ctx, cancel := context.WithTimeout(ctx, time.Second*50)
+	ctx, cancel := context.WithTimeout(s.ctx, time.Second*50)
```
cmars, line 326: Generally it's best to just pass the context through and let the caller decide how long they're willing to wait (or cancel the context with a signal handler, etc).

cmars (changes requested):
Recommend that we do not store the context in a struct. For more background on why, I can also recommend https://go.dev/blog/context-and-structs.

One reason to violate this general rule is for backwards compatibility, like http/Request.Context. Here we shouldn't be bound by compatibility though, as these are internal packages. Best to make a clean break IMO.

vervet-underground/internal/storage/gcs/client.go
genslein: Refactor makes this obsolete; removed

vervet-underground/internal/storage/gcs/client.go
genslein: Argh, my lack of go-ism...I've done the refactor here

cmars (approved):
Looks good now, thanks!

jcsackett (approved):
Looks good, thanks!

PR #201, opened 2022-05-09
chore: revert sts rotating credentials

genslein:
We don't need STS when in fact AWS Supports Role based environments for bare metal and EKS.

https://docs.aws.amazon.com/eks/latest/userguide/iam-roles-for-service-accounts-technical-overview.html


genslein:
related to https://github.com/snyk/vervet-underground/pull/22 messed up the naming on deploy for service account

genslein:
> LGTM. At some point can we document "how to set up & operate VU" in Backstage TechDocs?

Absolutely, we'll make a todo

PR #202, opened 2022-05-09
Fix: Handle success  on bucket found calls and fix idempotent get noise

genslein:
Found two bugs:

* we didn't correctly return on successful `HeadBucket` requests
* we didn't gracefully fail on 404 NotFound: NoSuchKey `GetObject` requests.

This has been successfully proven in pre-prod environment

### DIFF
--- a/vervet-underground/internal/storage/s3/client.go
+++ b/vervet-underground/internal/storage/s3/client.go
@@ -512,16 +512,8 @@ For information on other S3 API error codes see:
 https://aws.github.io/aws-sdk-go-v2/docs/handling-errors/
 */
 func handleAwsError(err error) error {
-	var opErr *smithy.OperationError
 	var apiErr smithy.APIError
 	var re *awshttp.ResponseError
-
-	if errors.As(err, &opErr) {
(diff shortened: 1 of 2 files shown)


davidharrigan (approved):
🚀

vervet-underground/internal/storage/s3/client.go
```diff
@@ -507,24 +512,28 @@ For information on other S3 API error codes see:
 https://aws.github.io/aws-sdk-go-v2/docs/handling-errors/
 */
 func handleAwsError(err error) error {
```
cmars, line 514: Do we have test coverage for this function? If not, can we add it?

cmars (approved):

vervet-underground/internal/storage/s3/client.go
genslein: I'll try to add something here before merge.

cmars: Log output might be captured by mocking the logger or patching `os.Stderr` to a bytes.Buffer (using https://pkg.go.dev/github.com/frankban/quicktest#C.Patch) and regex matching. LGTM without it though.

PR #206, opened 2022-06-21
feat: allow extending generate templates with Go functions

cmars:
This leaves generate.Generate open to extension with template functions
developed in Go.

Javascript template functions with the same name will take precedence
over Go functions provided through generate.GeneratorParams.

Built-in internal generator functions take precedence over both.


jcsackett (approved):
🎉 🎉 🎉

PR #208, opened 2022-06-24
fix: improve logging, fix s3 crash and inappropriate loader reuse

cmars:
Update vervet to latest version to better log collate errors by
identifying the documents which failed to merge.

Fix a panic in the s3 client that can occur when the client fails to
connect. Found this when trying to get localstack to work.

Fix inappropriate openapi3.Loader reuse among multiple documents. The
loader is very stateful and hangs on to references. This can be the
source of hard to isolate and debug problems in collate. It's possible
(though unknowable at this point) that this may clear up strange
conflicts that shouldn't be occuring in a live deployment with the GCS
backend.

### DIFF
--- a/vervet-underground/Dockerfile
+++ b/vervet-underground/Dockerfile
@@ -6,7 +6,7 @@ RUN go mod download
 RUN CGO_ENABLED=0 GOOS=linux go build -o server
 
 # Run
-FROM alpine:3.14
+FROM gcr.io/distroless/base-debian11
 WORKDIR /
 COPY --from=builder /src/server .
 EXPOSE 8080
--- a/vervet-underground/internal/storage/collator.go
+++ b/vervet-underground/internal/storage/collator.go
@@ -2,6 +2,7 @@ package storage
 
 import (
 	"fmt"
+	"net/url"
 	"sort"
 
 	"github.com/getkin/kin-openapi/openapi3"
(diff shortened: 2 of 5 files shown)


vervet-underground/Dockerfile
```diff
@@ -6,7 +6,7 @@ RUN go mod download
 RUN CGO_ENABLED=0 GOOS=linux go build -o server
 
 # Run
-FROM alpine:3.14
+FROM gcr.io/distroless/base-debian11
```
cmars, line 9: Drive-by security improvement.

vervet-underground/internal/storage/collator.go
```diff
@@ -96,18 +97,22 @@ func (c *Collator) Collate() (vervet.VersionSlice, map[vervet.Version]openapi3.T
...
 		if err != nil {
 			return nil, fmt.Errorf("could not load revision %s-%s-%s: %w", revision.Service, revision.Version, revision.Digest, err)
 		}
 
 		rv := &vervet.ResourceVersion{
-			Document: &vervet.Document{T: src},
-			Name:     revision.Service,
-			Version:  revision.Version,
+			Document: vervet.NewResolvedDocument(src, &url.URL{
+				Scheme: "vu",
```
cmars, line 110: This URL isn't used to resolve anything (the unregistered scheme guarantees that). It's used as an identifier in collate error messages, so we have something better than

`conflict in #/paths /examples/hello_world/{id}: declared in both  and ; conflict in #/paths /openapi: declared in both  and ; conflict in #/paths /openapi/{version}: declared in both  and `

vervet-underground/internal/storage/s3/client.go
```diff
@@ -537,12 +537,20 @@ func handleAwsError(err error) error {
 		fault = apiErr.ErrorFault().String()
 	}
 	if errors.As(err, &re) {
+		resp := re.HTTPResponse()
```
cmars, line 540: Based on the crashes I saw in compose, this can definitely be nil sometimes.

jcsackett (approved):
🚀

PR #210, opened 2022-06-27
fix: separate service name from URL, collate around openapi paths

cmars:
Two fixes:
- Separate the service name from the URL, so that endpoints can be
  updated without breaking historical continuity.
- Ignore /openapi endpoints during collation, if we've already merged
  these paths into the result.

### DIFF
--- a/vervet-underground/config/config_test.go
+++ b/vervet-underground/config/config_test.go
@@ -42,7 +42,7 @@ func TestLoad(t *testing.T) {
 	c.Run("specify config", func(c *qt.C) {
 		f := createTestFile(c, []byte(`{
 			"host": "0.0.0.0",
-			"services": ["localhost"]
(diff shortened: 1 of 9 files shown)


vervet-underground/config/config_test.go
```diff
@@ -176,11 +176,35 @@ func TestLoad(t *testing.T) {
...
+			"services": [{"url":"http://user:password@localhost"}],
+			"storage": {
+				"type": "memory"
+			}
+		}`))
+		_, err := config.Load(cfg.Name())
+		c.Assert(err, qt.ErrorMatches, `missing service name`)
+	})
+
+	c.Run("invalid service config - no name", func(c *qt.C) {
```
gablaxian, line 199: Same test name as above. Maybe
'invalid service config - name is a duplicate' ?

PR #211, opened 2022-06-27
fix: only collate configured services

### LINKED ISSUE #209: VU: cannot remove old services, cannot override paths
It's currently not possible to remove a service from VU aggregation. This is a problem if you break one service into two or more and want to aggregate them into a compiled API. The prior monolith will conflict forever with the new services because the storage implementations pull in all the old versions of the old service and it claims all the paths.

It's also not possible to selectively say:
* Aggregate everything from Old Monolith service except /path1, /path2 and /path3 which we moved to Shiny New Service
* Aggregate everything from Shiny New Service

It's also not possible to maintain historical continuity if the service scrape URL changes.

In order to be able to use VU in a dynamic services world, we need several things to happen:

- [ ] Stop collating services into the result aggregate specs that have been removed from the scrape list.
- [ ] Stop identifying services by their hostname in storage, but instead use a logical name. Logical name points at a hostname which can then change without disrupting history.
- [ ] Path matching patterns per service to limit what gets merge from each service. We probably need both include and exclude patterns for this...

cmars:
If a service is removed from configuration, stop collating it. Storage is left intact.

This fixes an issue where past service scrapes continue to show up in the collated output, even when removed from the services configuration.

The same result might be had by removing a service from config and manually deleting items from storage on deploy, but this is not a very desirable thing to do in production. It's akin to a stop-the-world manual database migration.

Stacked on #210 
Fixes #209

### DIFF
--- a/vervet-underground/internal/scraper/scraper.go
+++ b/vervet-underground/internal/scraper/scraper.go
@@ -164,7 +164,11 @@ func (s *Scraper) scrape(ctx context.Context, scrapeTime time.Time, svc service)
 }
 
 func (s *Scraper) collateVersions(ctx context.Context) error {
-	return s.storage.CollateVersions(ctx)
+	services := make([]string, len(s.services))
+	for i := range s.services {
+		services[i] = s.services[i].name
+	}
+	return s.storage.CollateVersions(ctx, services...)
 }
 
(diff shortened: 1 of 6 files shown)


cmars:
Superseded by #216

PR #215, opened 2022-07-11
feat: adds spec overlay functionality to VU

gablaxian:


cmars:
Still TODO on this branch: find a better way to pass overlays into config.

Since overlays are typically multiline YAML, passing them in config.json is going to be quite horrible to write and read.

Better alternative might be a separate CLI option like:

`vervet-underground --overlay overlay.yaml --config config.json`

That way we can write them as YAML in the helm configmap template...

PR #216, opened 2022-07-12
fix: CollateVersions filters out old services

jcsackett:
Previously vervet underground storage.Storage would collate versions
from all stored services, even if those services had been removed from
the configuration (e.g. because they were sunsetted).

Config now provides a ServiceFilter method which gets a filter of
current service names that is used in CollateVersions to remove any
versions from services *not* in the config, so old services are not
returned as part of the final openapi spec.

### DIFF
--- a/vervet-underground/internal/storage/gcs/client.go
+++ b/vervet-underground/internal/storage/gcs/client.go
@@ -117,7 +117,7 @@ func (s *Storage) NotifyVersions(ctx context.Context, name string, versions []st
 
 // CollateVersions iterates over all possible permutations of Service versions
 // to create a unified version spec for each unique vervet.Version.
-func (s *Storage) CollateVersions(ctx context.Context) error {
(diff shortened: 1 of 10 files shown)


vervet-underground/internal/storage/testing.go
```diff
@@ -0,0 +1,34 @@
+package storage
```
cmars, line 1: I'd put this in a separate `vervet-underground/internal/storage/testing` package.

vervet-underground/internal/storage/gcs/client.go
```diff
@@ -117,7 +117,7 @@ func (s *Storage) NotifyVersions(ctx context.Context, name string, versions []st
 
 // CollateVersions iterates over all possible permutations of Service versions
 // to create a unified version spec for each unique vervet.Version.
```
cmars, line 119: Might be nice to add information about the `serviceFilter` parameter to the godoc.

cmars (approved):

PR #223, opened 2022-07-18
fix: merge top-level OpenAPI extensions

jlourenc:
This PR implements appropriate merge of top-level OpenAPI extensions.

`x-snyk-api-stability` seems to be a special extension that needs to be removed from the merge for the tests to pass. I'd like to get some more eyes on this.

### DIFF
--- a/collator.go
+++ b/collator.go
@@ -228,22 +228,24 @@ func (c *Collator) Collate(rv *ResourceVersion) error {
 		return err
 	}
 
-	err = c.mergeComponents(rv)
-	if err != nil {
-		errs = multierr.Append(errs, err)
-	}
+	mergeExtensions(c.result, rv.T, false)
 	mergeInfo(c.result, rv.T, false)
-	err = c.mergePaths(rv)
-	if err != nil {
-		errs = multierr.Append(errs, err)
-	}
+	mergeOpenAPIVersion(c.result, rv.T, false)
 	mergeSecurityRequirements(c.result, rv.T, false)
 	mergeServers(c.result, rv.T, false)
-	err = c.mergeTags(rv)
-	if err != nil {
+
+	if err = c.mergeComponents(rv); err != nil {
 		errs = multierr.Append(errs, err)
 	}
-	mergeOpenAPIVersion(c.result, rv.T, false)
+
+	if err = c.mergePaths(rv); err != nil {
+		errs = multierr.Append(errs, err)
+	}
+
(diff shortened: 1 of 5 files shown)


merge.go
```diff
@@ -135,6 +136,23 @@ func mergeComponents(dst, src *openapi3.T, replace bool) {
...
 }
 
+func mergeExtensions(dst, src *openapi3.T, replace bool) {
+	if src.Extensions != nil && dst.Extensions == nil {
+		dst.Extensions = make(map[string]interface{}, len(src.Extensions))
+	}
+
+	for k, v := range src.Extensions {
+		// TODO: x-snyk-api-stability seems to be a special extension.
+		if k == ExtSnykApiStability {
```
jlourenc, line 146: How can I handle this better? I'm lacking understanding here. Help please ! 😄

cmars: This extension is used to indicate the stability level of an OpenAPI specification. It can't be directly merged because the effective API at a given stability level includes all operations (path + method) of equal or greater stability.

To illustrate, an experimental OpenAPI on a given version date also includes beta and GA operations effective on that version date. So while we merge or collate these together, we can't merge the stability level declared by the OpenAPI specs we're merging, or we'd misrepresent the stability of the resulting OpenAPI output.

So skipping this specific extension is reasonable.

collator.go
```diff
@@ -228,22 +228,24 @@ func (c *Collator) Collate(rv *ResourceVersion) error {
 		return err
 	}
 
-	err = c.mergeComponents(rv)
-	if err != nil {
+	mergeExtensions(c.result, rv.T, false)
```
cmars, line 231: Why are the merge calls in `Collate` re-ordered? Better alignment with the document structure?

merge.go
```diff
@@ -135,6 +136,23 @@ func mergeComponents(dst, src *openapi3.T, replace bool) {
 	}
 }
 
+func mergeExtensions(dst, src *openapi3.T, replace bool) {
+	if src.Extensions != nil && dst.Extensions == nil {
+		dst.Extensions = make(map[string]interface{}, len(src.Extensions))
+	}
+
+	for k, v := range src.Extensions {
+		// TODO: x-snyk-api-stability seems to be a special extension.
```
cmars, line 145: This comment should be updated with the explanation given below -- it's possible for specs to be merged from multiple stabilities and we don't want these different stability inputs to override the declared stability of the output we're building.

PR #228, opened 2022-07-22
fix(VU): Reduce GCS storage locking and test improvements

cmars:
Minimize the lock scope in the GCS storage backend to protect only
access to state read/written concurrently (collatedVersions).

Restructure GCS storage test setup to drive a fake GCS service with
testcontainers-go and use this instance for testing in local development
and CI.


jcsackett (approved):
Looks good, @cmars!

PR #240, opened 2022-09-05
Test/duplicate spec files

juliabanerjee:


resource.go
```diff
@@ -183,11 +183,18 @@ func (e resourceVersionSlice) Swap(i, j int) { e[i], e[j] = e[j], e[i] }
...
 // document.
 func LoadResourceVersions(epPath string) (*ResourceVersions, error) {
+	// Partially handles edge case where there is a spec.yml and spec.yaml file for the same API
+	// It is assumed that duplicate specs would cause an error elsewhere in vervet
 	specYamls, err := filepath.Glob(epPath + "/*/spec.yaml")
 	if err != nil {
 		return nil, err
 	}
-	return LoadResourceVersionsFileset(specYamls)
+	specYmls, err := filepath.Glob(epPath + "/*/spec.yml")
```
gablaxian, line 192: I took a quick look into `Glob` and I'm wondering now that, as I heard the pattern `spec.{yaml, yml}` doesn't work, does this instead:
`spec.ya?ml`? We may still be able to simplify the logic.

juliabanerjee: I tried that before but it didn't work because the function is looking for a match to the specific string given. Therefore, it would be looking for a file matching the name "spec.ya?ml" rather than a spec.yaml and/or a spec.yml

cmars: Suggest the use of github.com/bmatcuk/doublestar for globbing here, which [supports patterns](https://github.com/bmatcuk/doublestar#patterns) like `{yaml,yml}`. doublestar is already a vervet dependency.

juliabanerjee: Thank you, I will give that a go

gablaxian (approved):

resource_test.go
```diff
@@ -101,3 +101,10 @@ func TestIsExtensionNotFound(t *testing.T) {
 	_, err = ExtensionString(resource.ExtensionProps, ExtSnykApiStability)
 	c.Assert(IsExtensionNotFound(err), qt.IsFalse)
 }
+
+func TestLoadResourceVersionsWithDuplicateSpecs(t *testing.T) {
+	c := qt.New(t)
+	ep, err := LoadResourceVersions(testdata.Path("duplicate-specs"))
+	c.Assert(ep, qt.IsNotNil)
+	c.Assert(err, qt.IsNil)
```
cmars, line 109: I think this should result in an error, if there are duplicate specs in the same directory.

resource.go
```diff
@@ -183,11 +184,13 @@ func (e resourceVersionSlice) Swap(i, j int) { e[i], e[j] = e[j], e[i] }
 // ExtSnykApiStability extension value at the top-level of the OpenAPI
 // document.
 func LoadResourceVersions(epPath string) (*ResourceVersions, error) {
-	specYamls, err := filepath.Glob(epPath + "/*/spec.yaml")
+	// Handles case where there is either a spec.yml or spec.yaml file but not edge case where there are both specs for the same API
+	// It is assumed that duplicate specs would cause an error elsewhere in vervet
```
cmars, line 188: I disagree with this assumption. I think we should detect and error on duplicates (`spec.yaml` and `spec.yml`) here, rather than place the burden of de-duplicating on the caller.

cmars (changes requested):

cmars (approved):
Looks good now, thanks for this!

PR #244, opened 2022-09-08
fix: separated the lifecycle date into deprecation and release

carissadean:
### What this does
Separates the lifecycle date into two dates used for deprecation and release.

- For determining deprecation we use the `VERVET_LIFECYCLE_AT` env var, and if not set default to the current date.
- For determining whether a version is unreleased we always use the current date


cmars (approved):
Looks great, thanks for this!

PR #245, opened 2022-09-14
Fix/collate conflict bug

carissadean:
Sorts revisions by timestamp, with the most recent revision being first.

More info: https://snyksec.atlassian.net/browse/ELK-709

### DIFF
--- a/vervet-underground/internal/storage/revision.go
+++ b/vervet-underground/internal/storage/revision.go
@@ -27,7 +27,7 @@ type ContentRevision struct {
 // ServiceRevisions tracks a collection of ContentRevisions and API uniqueVersions for a single service.
 type ServiceRevisions struct {
 	// Revisions is a map of version to a collection of revisions.  During collation, content revision with the latest scraping timestamp is used.
-	Revisions map[vervet.Version][]ContentRevision
+	Revisions map[vervet.Version]Revisions
 	// Versions is a collection of API uniqueVersions that this service serves.
 	Versions vervet.VersionSlice
 }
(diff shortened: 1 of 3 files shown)


vervet-underground/internal/storage/revision.go
```diff
@@ -75,3 +75,16 @@ func (s ServiceRevisions) ResolveLatestRevision(version vervet.Version) (Content
 	}
 	return revision, nil
 }
+
+type Revisions []ContentRevision
```
cmars, line 79: Suggest naming this `ContentRevisions` for a clearer association with the slice element type.

vervet-underground/internal/storage/revision.go
```diff
@@ -75,3 +75,16 @@ func (s ServiceRevisions) ResolveLatestRevision(version vervet.Version) (Content
...
+
+// Less implements sort.Interface.
+func (r Revisions) Less(i, j int) bool {
+	return r[i].Timestamp.After(r[j].Timestamp)
```
cmars, line 83: Suggest making the ordering deterministic even when timestamps aren't unique. The `Digest` would be a good tie-breaker if the timestamps are equal.

Seems like a remote edge case, but time sometimes runs strangely within VMs and containers.

cmars (approved):
LGTM with a couple of suggestions. I may pick these up this afternoon & land if you're OK with it.

carissadean:
> LGTM with a couple of suggestions. I may pick these up this afternoon & land if you're OK with it.

Yes feel free! @cmars

PR #249, opened 2022-09-26
fix: resolve OpenAPI validators at the correct stability level

### LINKED ISSUE #250: versionSlice Resolve does not resolve to an appropriate version.
Adding the below test case to `TestVersionSlice`:
```
{
    versions: VersionSlice{
        MustParseVersion("2021-09-06~beta"),
	MustParseVersion("2021-09-07"),
	MustParseVersion("2021-09-08~beta"),
	MustParseVersion("2021-09-09~experimental"),
    },
    first: "2021-09-06~beta",
    last:  "2021-09-09~experimental",
    matchTests: []matchTest{{
        match:  "2021-09-10",
	result: "2021-09-07",
    }},
}
```
yields to a `no matching version` error.

cmars:
Related to snyk/vervet#248, a similar stability version issue affects
validation middleware "versionware".

This breaking change replaces VersionSlice.Resolve with
VersionIndex.Resolve.

It turns out resolving versions correctly across date and stability
requires tracking the most recent release per stability for each
distinct version date. This requires a different structure for indexing;
a simple []Version slice does not suffice for an O(logN) search
complexity.

Fixes #250.

### DIFF
--- a/spec.go
+++ b/spec.go
@@ -20,7 +20,8 @@ const SpecGlobPattern = "**/[0-9][0-9][0-9][0-9]-[0-9][0-9]-[0-9][0-9]/spec.yaml
 // SpecVersions stores a collection of versioned OpenAPI specs.
 type SpecVersions struct {
 	versions  VersionSlice
-	documents []*openapi3.T
+	index     VersionIndex
+	documents map[Version]*openapi3.T
 }
 
 // LoadSpecVersions returns SpecVersions loaded from a directory structure
@@ -76,21 +77,15 @@ func (sv *SpecVersions) Versions() VersionSlice {
 // returned. Returns ErrNoMatchingVersion if there is no release matching this
 // version.
 func (sv *SpecVersions) At(v Version) (*openapi3.T, error) {
-	i, err := sv.versions.ResolveIndex(v)
+	resolvedVersion, err := sv.index.Resolve(v)
 	if err != nil {
 		return nil, err
 	}
-	// Because this collection contains each distinct version date and
-	// stability, perform an exact match on the most recent version date
-	// matching the requested stability. ResolveIndex may overshoot it because
-	// it returns equal or greater stability.
-	for ; i >= 0; i-- {
-		checkVersion := sv.versions[i]
-		if dateCmp, stabilityCmp := checkVersion.compareDateStability(&v); dateCmp <= 0 && stabilityCmp == 0 {
-			return sv.documents[i], nil
-		}
+	doc, ok := sv.documents[resolvedVersion]
+	if !ok {
+		panic(fmt.Sprintf("missing expected document for version %v", resolvedVersion))
 	}
-	return nil, ErrNoMatchingVersion
+	return doc, nil
 }
 
 func (sv *SpecVersions) resolveOperations() error {
@@ -109,8 +104,8 @@ func (sv *SpecVersions) resolveOperations() error {
 	}
 	type operationVersionMap map[operationKey]operationVersion
 	activeOpsByStability := map[Stability]operationVersionMap{}
-	for i, v := range sv.versions {
-		doc := sv.documents[i]
+	for _, v := range sv.versions {
+		doc := sv.documents[v]
 		currentActiveOps, ok := activeOpsByStability[v.Stability]
 		if !ok {
 			currentActiveOps = operationVersionMap{}
@@ -237,6 +232,10 @@ func newSpecVersions(specs resourceVersionsSlice) (*SpecVersions, error) {
 			} else if err != nil {
 				return nil, err
 			}
+			if doc.ExtensionProps.Extensions == nil {
+				doc.ExtensionProps.Extensions = map[string]interface{}{}
+			}
+			doc.ExtensionProps.Extensions[ExtSnykApiVersion] = v.String()
 			documentVersions[v] = doc
 		}
 	}
@@ -247,14 +246,8 @@ func newSpecVersions(specs resourceVersionsSlice) (*SpecVersions, error) {
 	sort.Sort(versions)
 	sv := &SpecVersions{
 		versions:  versions,
-		documents: make([]*openapi3.T, len(versions)),
-	}
-	for i := range versions {
-		sv.documents[i] = documentVersions[versions[i]]
-		if sv.documents[i].ExtensionProps.Extensions == nil {
-			sv.documents[i].ExtensionProps.Extensions = map[string]interface{}{}
-		}
-		sv.documents[i].ExtensionProps.Extensions[ExtSnykApiVersion] = versions[i].String()
+		index:     NewVersionIndex(versions),
+		documents: documentVersions,
 	}
 	err := sv.resolveOperations()
 	if err != nil {
--- a/version.go
+++ b/version.go
@@ -4,6 +4,7 @@ package vervet
 import (
 	"fmt"
 	"os"
+	"sort"
 	"strings"
 	"time"
 )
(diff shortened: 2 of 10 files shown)


versionware/validator_resolve_test.go
```diff
@@ -0,0 +1,109 @@
...
+	}, {
+		url:    "/foo?color=red&version=2022-07-20~beta",
+		status: 200,
+	}, {
+		url:    "/foo?flavor=flav&version=2022-07-21",
+		status: 200,
+	}, {
+		url:    "/foo?version=2022-01-01",
+		status: 404,
+	}}
```
jcsackett, line 91: It'll be a lot easier to sort out what might be going wrong here if we had test names for the cases; e.g. "no resolved version", "resolved exact match" etc.

Not necessary, but a suggestion.

jcsackett (approved):

cmars:
@jlourenc found a problem with `VersionSlice.Resolve`. Looking into it, we found that resolving the correct stability version requires a different data structure. Rather than add it alongside a broken search implementation, I decided to fix it in a new major rev.

version.go
```diff
@@ -227,75 +241,108 @@ func VersionDateStrings(vs []Version) []string {
...
-		return Version{}, err
+// VersionIndex provides a search over versions, resolving which version is in
+// effect for a given date and stability level.
+type VersionIndex struct {
+	versions []effectiveVersion
+}
+
+type effectiveVersion struct {
+	date        time.Time
+	stabilities map[Stability]time.Time
```
jlourenc, line 255: nit: Using the `map` here is totally fine. Just bare in mind that `Stability` is defined as integer. More than that the integers are consecutive. So it could be defined as an array (not a slice) which gives you fast lookup, allows you to iterate through elements in (reverse) order.

version.go
```diff
@@ -227,75 +241,108 @@ func VersionDateStrings(vs []Version) []string {
...
+func NewVersionIndex(vs VersionSlice) *VersionIndex {
+	sort.Sort(vs)
+	var evs []effectiveVersion
+	evIndex := -1
+	currentStabilities := map[Stability]time.Time{}
+	for i := range vs {
+		if evIndex == -1 || !evs[evIndex].date.Equal(vs[i].Date) {
+			evs = append(evs, effectiveVersion{
+				date:        vs[i].Date,
+				stabilities: map[Stability]time.Time{},
```
jlourenc, line 269: You could do `stabilities: maps.Clone(currentStabilities)` if you're willing to upgrade the minimum Go version supported by vervet to `1.17` which is now unsupported by the Go team since the release of Go 1.19.

https://pkg.go.dev/golang.org/x/exp@v0.0.0-20220927162542-c76eaa363f9d/maps#Clone

version.go
```diff
@@ -227,75 +241,108 @@ func VersionDateStrings(vs []Version) []string {
...
-// ResolveIndex returns the slice index of the most recent Version in the slice
-// with equal or greater stability.
-//
-// This method requires that the VersionSlice has already been sorted with
-// sort.Sort, otherwise behavior is undefined.
-func (vs VersionSlice) ResolveIndex(q Version) (int, error) {
-	lower, curr, upper := 0, len(vs)/2, len(vs)
+// resolveIndex performs a binary search on the stability versions in effect on
+// the query date.
+func (vi *VersionIndex) resolveIndex(query time.Time) (int, error) {
```
jlourenc, line 286: I think this function can be simplified to:
```
func (vi VersionIndex) resolveIndex(query time.Time) (int, error) {
        if len(vi.versions) == 0 || vi.versions[0].date.After(query) {
                // empty slice or oldest version more recent than query
		return -1, ErrNoMatchingVersion
	}

	lower, curr, upper := 0, len(vi.versions)/2, len(vi.versions)
	for lower < upper-1 {
		if vi.versions[curr].date.After(query) {
			upper = curr
		} else {
			lower = curr
		}
		curr = lower + (upper-lower)/2
	}

	return lower, nil
}
```

version.go
```diff
@@ -227,75 +241,108 @@ func VersionDateStrings(vs []Version) []string {
...
+	i, err := vi.resolveIndex(query.Date)
+	if err != nil {
+		return Version{}, err
+	}
+	for stab, ok := query.Stability, true; ok; stab, ok = stab.Next() {
+		matchDate, ok := vi.versions[i].stabilities[stab]
+		if ok {
+			return Version{Date: matchDate, Stability: stab}, nil
+		}
+	}
```
jlourenc, line 324: I think this introduces a slightly different behaviour than before. It returns the "lowest" stability available that is equal or greater of the query's stability. Previously, it would have returned the greatest stability available.

version_test.go
```diff
@@ -143,13 +142,19 @@ func TestVersionSlice(t *testing.T) {
...
+		}, {
+			match:  "2021-06-11~beta",
+			result: "2021-06-10~beta",
+		}, {
+			match:  "2021-06-11",
 			result: "2021-06-10",
 		}, {
 			match:  "2021-06-10~experimental",
-			result: "2021-06-10",
+			result: "2021-06-10~beta",
```
jlourenc, line 154: I guess this is the change of behaviour I've mentioned just above. It's great to see our tests have captured that 🎉

spec.go
```diff
@@ -76,21 +77,15 @@ func (sv *SpecVersions) Versions() VersionSlice {
...
-	// matching the requested stability. ResolveIndex may overshoot it because
-	// it returns equal or greater stability.
-	for ; i >= 0; i-- {
-		checkVersion := sv.versions[i]
-		if dateCmp, stabilityCmp := checkVersion.compareDateStability(&v); dateCmp <= 0 && stabilityCmp == 0 {
-			return sv.documents[i], nil
-		}
+	doc, ok := sv.documents[resolvedVersion]
+	if !ok {
+		return nil, ErrNoMatchingVersion
```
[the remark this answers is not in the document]
cmars: Agree, such a case would be caused by a regression in either `newSpecVersions` or `NewVersionIndex`. A panic seems more appropriate here.

version.go
cmars: Much nicer, thanks!

version.go
cmars: Stability levels have become well-enough defined by this point that I'm ready to fully embrace their inner int nature. Thanks for the nudge here :)

version.go
cmars: Going with the slice approach.

version.go
cmars: Good callout here. It turns out there are two different kinds of "resolving" needed, which answer two different questions:

* I want to interact with a _service's API_ at a given stability on a given date. The versions have already been compiled from resources such that a version date+stability exists in my index for each available distinct API release. What is the best match?
* I want to comple a service-level API spec from many resources. When looking at each resource, what is the most stable API available from that _resource_ on a given date, greater than or equal to a minimum stability? In this scenario, a resource may have only one release per day (due to the directory structure used for expressing resource versions).

The first case is the one most will care about -- this is what you would typically use to decide what version of a service you want to interact with, what version of the service's OpenAPI spec should be used to validate a request for an arbitrary version, for example.

The second case is used only when compiling a service's OpenAPI specs.

version_test.go
cmars: Indeed. I've expanded this test to provide coverage for both `Resolve` and `resolveForBuild`. There wasn't a clear enough distinction between the two prior to this PR, which was really at the root of the incorrect resolving behavior.

jcsackett (approved):
New changes look good, thanks!

PR #259, opened 2022-12-07
fix: merge specs instead of overwriting

jgresty:
If an api config declared resources from multiple paths then `build` would iterate over them and overwrite the previous iteration with the latest. This would mean that generated specs would only have resources from the last path specified.

This patch merges all resources before writing so that all paths are considered.


genslein:
@jlourenc #270 #269 should supercede this PR correct? Can we close this?

jlourenc:
I believe this is a different use case. There's value in pursuing this fix with a rebase and addition of tests.

PR #288, opened 2023-04-12
fix: pick newer versions of endpoints when collating over services

jgresty:
If an endpoint is moved from one service to another, then there is a chance that we will continue to use the path from the old service even though the newer one has a more recent version.

It is assumed that all services are independent when doing version resolution, then all the services are merged together. This is fine if the services do not have overlapping paths but will cause unexpected behaviour otherwise.

We keep the first spec for each endpoint we find when collating versions, which means we need to find the most recent version first. To do this we can change the order we process specs by using the vervet version before the scrape timestamp - which will mean more recent versions will be processed first.

### DIFF
--- a/collator_test.go
+++ b/collator_test.go
@@ -172,3 +172,26 @@ func TestCollatePathConflict(t *testing.T) {
 	c.Assert(err, qt.ErrorMatches, `.*conflict in #/paths /examples/hello-world/{id2}: declared in both.*`)
 	c.Assert(err, qt.ErrorMatches, `.*conflict in #/paths /examples/hello-world: declared in both.*`)
 }
+
+func TestCollateMergingResources(t *testing.T) {
+	c := qt.New(t)
+	collator := vervet.NewCollator(vervet.UseFirstRoute(true))
+
+	newService, err := vervet.LoadResourceVersions(testdata.Path("competing-specs/special_projects"))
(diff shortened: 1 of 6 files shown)


collator_test.go
```diff
@@ -172,3 +172,26 @@ func TestCollatePathConflict(t *testing.T) {
...
+	specV2, err := originalService.At("2021-08-20~experimental")
+	c.Assert(err, qt.IsNil)
+
+	err = collator.Collate(specV2)
+	c.Assert(err, qt.IsNil)
+	err = collator.Collate(specV1)
+	c.Assert(err, qt.IsNil)
+
+	result := collator.Result()
+	c.Assert(result.Paths["/orgs/{org_id}/projects/{project_id}"].Delete.Responses["204"], qt.IsNotNil)
```
cmars, line 196: Should we assert some feature of the delete operation here, to prove that we're using the new operation. Maybe that `focus_type` property?

ajbeattie-snyk (approved):
LGTM as far as I can understand from limited context (mainly the slack thread). I'll defer to Casey on any finer points, but adding a ✅ from extensibility so you can merge when ready.

cmars (approved):
LGTM

collator_test.go
jgresty: The old path only responds with 200, not 204 so this is asserting that

PR #289, opened 2023-04-17
chore: update build file

gablaxian:


jgresty (approved):
It is surprising we did not catch this in CI, looks like we aren't even running the tests in this directory...

love-bhardwaj (approved):
Suggestion/nit: add makefile tidy command that updates both the root module and this 🙂

PR #291, opened 2023-05-01
feat: skip saving legacy version when running scrapper

clwiseman:
This change will enable the use of legacy APIs using a default version `2021-01-01` (which predates `vervet`) without `vervet-underground` scraping for the OpenAPI contract. This enables `vervet` to be used on legacy APIs which do not meet the OpenAPI standard.

To test to end to end:
- Run a service locally with an API using the default version `2021-01-01`.
- Used a `config.local.json` file to point to that service.
- Started `vervet-underground` after updating the `docker-compose.yml` and `server.go` to point to the `config.local.json` file
- Called `http://localhost:8080/openapi` confirm the default version is missing from the response.

### DIFF
--- a/vervet-underground/internal/scraper/scraper.go
+++ b/vervet-underground/internal/scraper/scraper.go
@@ -155,6 +155,12 @@ func (s *Scraper) scrape(ctx context.Context, scrapeTime time.Time, svc service)
 	for i := range versions {
 		// TODO: we might run this concurrently per live service pod if/when
 		// we're more k8s aware, but we won't do that yet.
+
+		// Skip if it's a legacy api using the default legacy version.
+		if isLegacyVersion(versions[i]) {
+			continue
+		}
+
 		contents, isNew, err := s.getNewVersion(ctx, svc, versions[i])
 		if err != nil {
 			return errors.WithStack(err)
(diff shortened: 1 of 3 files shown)


genslein (approved):
just tag @snyk/extensibility for :shipit: rights

vervet-underground/internal/scraper/scraper.go
```diff
@@ -275,6 +281,12 @@ func (s *Scraper) hasNewVersion(ctx context.Context, svc service, version string
 	return s.storage.HasVersion(ctx, svc.name, version, digest)
 }
 
+// isLegacyVersion is used to identify legacy APIs which should be excluded.
+func isLegacyVersion(version string) bool {
+	// This default version predates vervet's creation date.
+	return version == "2021-01-01"
```
ajbeattie-snyk, line 287: Not sure if this is just a temporary measure, but we could make this a configurable cutoff date (or list of legacy versions) if we wanted to be more generic. 

That said, I'm not opposed to landing this version for now.

ajbeattie-snyk (approved):
LGTM, added a thought on making it a bit more generic if it's not an urgent blocker :)

vervet-underground/internal/scraper/scraper.go
clwiseman: We considered making this configurable to be more generic. The concern is if it's configured per-service, some of our other tooling could not keep track. By enforcing a single date we are being strict with what is considered a "legacy" API.

I am hoping this will be a temporary measure, but if it turns out to be a helpful feature for others then I am all for making it more generic 👍

PR #292, opened 2023-05-02
Chore/update golangci lint

love-bhardwaj:
## What this PR does?

Adds linting and formatting with formatted code. This should not alter the functionality of the code.


internal/backstage/backstage.go
```diff
@@ -88,7 +88,7 @@ type CatalogInfo struct {
 func (c *CatalogInfo) Save(w io.Writer) error {
 	enc := yaml.NewEncoder(w)
 	enc.SetIndent(2)
-	var docs []*yaml.Node
+	docs := []*yaml.Node{}
```
cmars, line 91: I'm curious what the purpose of this linter rule is. Is there a significant difference between a nil and empty slice? I don't have a strong opinion either way, just wondering why out loud.

internal/cmd/scaffold.go
```diff
@@ -51,7 +52,7 @@ func ScaffoldInit(ctx *cli.Context) error {
 		return err
 	}
 	err = sc.Organize()
-	if err == scaffold.ErrAlreadyInitialized {
+	if errors.Is(err, scaffold.ErrAlreadyInitialized) {
```
cmars, line 55: Let's make sure we have a test here as well.

internal/cmd/scaffold_test.go
cmars, line 1: Do we still need the `scaffold` command? I thought it was superseded by sweater-comb, and is unlikely to be maintained.

internal/compiler/compiler.go
```diff
@@ -286,7 +287,7 @@ func (c *Compiler) Build(ctx context.Context, apiName string) error {
 		versions := specVersions.Versions()
 		for _, version := range versions {
 			spec, err := specVersions.At(version)
-			if err == vervet.ErrNoMatchingVersion {
+			if errors.Is(err, vervet.ErrNoMatchingVersion) {
```
cmars, line 290: This needs test coverage.

internal/linter/optic/linter.go
```diff
@@ -180,7 +181,7 @@ func (o *Optic) Run(ctx context.Context, root string, paths ...string) error {
 	}
 	for i := range paths {
 		comparison, volumeArgs, err := o.newComparison(paths[i], fromFilter, toFilter)
-		if err == errHasException {
+		if errors.Is(err, errHasException) {
```
cmars, line 184: Test coverage...

resource.go
```diff
@@ -65,26 +65,26 @@ type extensionNotFoundError struct {
 }
 
 // Error implements error.
-func (e *extensionNotFoundError) Error() string {
-	return fmt.Sprintf("extension \"%s\" not found", e.extension)
+func (rv *extensionNotFoundError) Error() string {
```
cmars, line 68: I don't think `rv` is a good name for this method target. That seems more appropriate for the `ResourceVersion` methods below. Did the linter object to `e` here?

resource_versions.go
```diff
@@ -15,7 +16,7 @@ func (s resourceVersionsSlice) validate() error {
 		resourcePaths := map[string]string{}
 		for _, eps := range s {
 			ep, err := eps.At(v.String())
-			if err == ErrNoMatchingVersion {
+			if errors.Is(err, ErrNoMatchingVersion) {
```
cmars, line 19: Test coverage

resource_versions.go
```diff
@@ -50,7 +51,7 @@ func (s resourceVersionsSlice) at(v Version) (*openapi3.T, error) {
 	coll := NewCollator()
 	for _, eps := range s {
 		ep, err := eps.At(v.String())
-		if err == ErrNoMatchingVersion {
+		if errors.Is(err, ErrNoMatchingVersion) {
```
cmars, line 54: Test coverage

spec.go
```diff
@@ -227,7 +227,7 @@ func newSpecVersions(specs resourceVersionsSlice) (*SpecVersions, error) {
 		for _, stability := range stabilities {
 			v := Version{Date: date, Stability: stability}
 			doc, err := specs.at(v)
-			if err == ErrNoMatchingVersion {
+			if errors.Is(err, ErrNoMatchingVersion) {
```
cmars, line 230: Test coverage

testdata/output/embed.go
```diff
@@ -18,5 +18,6 @@ import "embed"
 //go:embed 2021-08-20~experimental/spec.yaml
 //go:embed 2021-08-20~beta/spec.json
 //go:embed 2021-08-20~beta/spec.yaml
+
```
cmars, line 21: Is this generated code? If so...

- Did we regenerate it?
- Should it have a "generated code do not edit" kind of comment at the top?

version.go
```diff
@@ -78,8 +79,11 @@ func (s Stability) String() string {
 		return "beta"
 	case StabilityGA:
 		return "ga"
+	case stabilityUndefined:
+		fallthrough
```
cmars, line 83: Not a fan of fallthrough... if it was me, I'd remove this `stabilityUndefined` case and let the `default` catch it.

version.go
```diff
@@ -199,8 +203,11 @@ func (v Version) Sunset(vr Version) (time.Time, bool) {
 		return vr.Date.Add(SunsetBeta), true
 	case StabilityGA:
 		return vr.Date.Add(SunsetGA), true
+	case stabilityUndefined:
+		fallthrough
```
cmars, line 207: Same here (but I won't block on it)

cmars (approved):
LGTM! I'd recommend checking for test coverage over all the error equality replacements with `errors.Is` -- or at least convince yourself that errors.Is will have the same behavior for these particular types of errors. If they're string-based which I think they all are, should be fine, but that's the riskiest change I found here.

Thanks for doing this! Getting this under control with a linter is something I'd wanted to do for a while.

internal/backstage/backstage.go
```diff
@@ -128,7 +128,7 @@ func LoadCatalogInfo(r io.Reader) (*CatalogInfo, error) {
 	for {
 		var node yaml.Node
 		err := dec.Decode(&node)
-		if err == io.EOF {
+		if errors.Is(err, io.EOF) {
```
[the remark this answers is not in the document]
love-bhardwaj: I plan to remove the linter in charge of this task temporarily while we work on creating test cases to address these issues. Once the test cases are in place, we can reintroduce the linter.

internal/cmd/scaffold_test.go
love-bhardwaj: I don't have much experience with "vervet," so I'm not certain if it's still supported or not.

testdata/output/embed.go
love-bhardwaj: I am ➕ on adding the top comment with something like "generated code do not edit".

It was regenerated because it was conflicting with general go formatting rules.

internal/backstage/backstage.go
love-bhardwaj: I am used to pre-allocating instead of using `var` but that's just me 😅 I think it shouldn't affect much which we use but the linter author sites performance as the reason for this. Ref: https://github.com/alexkohler/prealloc

PR #294, opened 2023-05-03
chore: consistent go and golangci-lint versions for both vervet and VU

love-bhardwaj:
- Update the vervet underground go and `golangci-lint` versions to match that of `vervet` to ensure a consistent experience.
- Update the `vervet undergound` `makefile` to install `golangci-lint` in the local bin, enabling us to work on `vervet underground` independently. It also replicates the setup of `vervet`.
- Update docker and ci config files to use the correct version of go and `golangci-lint`.

### DIFF
--- a/vervet-underground/Makefile
+++ b/vervet-underground/Makefile
@@ -1,3 +1,9 @@
+GO_BIN=$(shell pwd)/.bin/go
+
+SHELL:=env PATH=$(GO_BIN):$(PATH) $(SHELL)
+
+GOCI_LINT_V?=v1.51.0
+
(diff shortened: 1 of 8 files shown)


vervet-underground/Makefile
```diff
@@ -8,15 +14,15 @@ clean:
 
 .PHONY: lint
 lint:
-	golangci-lint run  --fix -v ./...
+	../.bin/go/golangci-lint run  --fix -v ./...
```
jgresty, line 17: Why has this changed?

jgresty (approved):

PR #300, opened 2023-06-14
feat: prevent sunset specs being compiled into new versions

mikedevelops:
Prevent resources that have been explicitly marked as "sunset" from being compiled into newer versions of a service's API.

More detailed information on the approach here: https://www.notion.so/snyk/Sunsetting-OpenAPI-specs-with-Vervet-d7397f98a42c45a683a7e6296d1b4d5a?pvs=4 

☝️ If we can agree on the approach internally, documentation will be added to reflect the change in behavior.


resource.go
```diff
@@ -159,6 +159,13 @@ func (rv *ResourceVersions) At(vs string) (*ResourceVersion, error) {
 	if !ok {
 		return nil, ErrNoMatchingVersion
 	}
+
+	// skip resolving versions for resources that have been marked as sunset in a previous version
+	if lifecycle, err := r.Document.Lifecycle(); err == nil && lifecycle == LifecycleSunset &&
```
DarrellMozingo, line 164: Ah, are we putting sunset at the top level as well as here because resolution isn't an operation specific job?

DarrellMozingo (approved):
Nice one 👏

resource.go
mikedevelops: Yeah exactly. As an improvement, we might want to do it the other way around, i.e. mark the resource version as sunset at the top level so it isn't collated, and at the same time have Vervet add the per-operation sunset annotations so we can prevent the promotion of operations from previous versions (which is done later at the operation level).

For now, the path of least resistance is to be explicit in the spec and mark everything as sunset.

jgresty (approved):
Nice, I'm sure this is going to break something for someone with the amount of experimental apis we have in flight so we might want to announce something when merging

jcsackett (approved):
Thanks, this looks great!

mikedevelops:
> Nice, I'm sure this is going to break something for someone with the amount of experimental apis we have in flight so we might want to announce something when merging

The behaviour of not promoting sunset operations should only take affect for resource APIs that have been explicitly marked sunset in the src. This doesn't support resources that are auto-sunset – this is intentional for now becaues we don't want to break everything as you mention.

I think the place we want to get to is consistency and have auto-sunset resources behave the same way, but I'd rather create a smaller footprint on this quick fix.

PR #303, opened 2023-08-31
feat: publish orb for vervet-underground

VinceDeslo:
This PR migrates publishing of vervet-underground to use Snyk's publish orb.
This ensures that the container image gets scanned.


VinceDeslo:
Duplicate that was replaced by a rebased branch

PR #304, opened 2023-09-01
feat: add publish orb for vu

VinceDeslo:
This PR updates publishing of vervet-underground to use Snyk's internal publish orb.
This ensures that the container image gets scanned.

### DIFF
--- a/testdata/catalog-vervet-apis.yaml
+++ b/testdata/catalog-vervet-apis.yaml
@@ -192,15 +192,15 @@ metadata:
     api.snyk.io/generated-by: vervet
   labels:
     api.snyk.io/version-date: "2023-06-01"
-    api.snyk.io/version-lifecycle: deprecated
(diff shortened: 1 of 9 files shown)


jgresty (approved):

testdata/catalog-vervet-apis.yaml
love-bhardwaj, line 1: Any reason why things in here would change?

love-bhardwaj (approved):

testdata/catalog-vervet-apis.yaml
VinceDeslo: The [backstage_test.go](https://github.com/snyk/vervet/blob/feat/EXT-2052-add-vu-container-scanning/internal/backstage/backstage_test.go) tests rely on this file as a fixture, yet it doesn't get updated with any go generate commands.

Since some of the testdata out files were automatically sunset after 90 days, this fixture needed to be updated to reflect that.

PR #313, opened 2023-11-03
chore(vu): upgrade Go deps

jlourenc:
Update vervet-underground Go dependencies and adapt the code to breaking changes introduced in Vervet v5.


vervet-underground/internal/storage/revision.go
```diff
@@ -26,29 +26,32 @@ type ContentRevision struct {
...
-	Revisions map[vervet.Version]ContentRevisions
-	// Versions is a collection of API uniqueVersions that this service serves.
-	Versions vervet.VersionSlice
+	// revisions is a map of version to a collection of revisions.  During collation, content revision with the latest scraping timestamp is used.
+	revisions map[vervet.Version]ContentRevisions
+	// versions is a collection of API uniqueVersions that this service serves.
+	versions vervet.VersionSlice
+	// versionIndex is an index of API uniqueVersion that this service serves.
+	versionIndex     vervet.VersionIndex
+	generateNewIndex bool
```
ajbeattie-snyk, line 35: Nit: can you add a comment explaining the purpose/behavior of `generateNewIndex`?

vervet-underground/internal/storage/revision.go
```diff
@@ -26,29 +26,32 @@ type ContentRevision struct {
...
+		generateNewIndex: true,
 	}
 }
 
 // Add registers a new ContentRevision for the service.
 func (s *ServiceRevisions) Add(revision ContentRevision) {
 	version := revision.Version
-	if _, ok := s.Revisions[version]; !ok {
-		s.Versions = append(s.Versions, version)
-		sort.Sort(s.Versions)
```
ajbeattie-snyk, line 48: Just to double check - does `s.versions` not need to be sorted? (not very familiar with this so unsure if that's necessary)

vervet-underground/internal/storage/revision.go
```diff
@@ -26,29 +26,32 @@ type ContentRevision struct {
...
+	if _, ok := s.revisions[version]; !ok {
+		s.versions = append(s.versions, version)
+		s.generateNewIndex = true
```
ajbeattie-snyk, line 51: Would it make sense to generate the new index here for simplicity (and remove `generateNewIndex`? Again, sorry if I'm missing some context that is not touched in the PR 😅

ajbeattie-snyk (approved):
LGTM in general, added a few questions mainly to be sure on things I'm not familiar with!

vervet-underground/internal/storage/revision.go
jlourenc: It does when doing resolution. This is now handled by `vervet.VersionIndex` which sorts the `vervet.VersionSlice` provided on init.

vervet-underground/internal/storage/revision.go
jlourenc: Yes, it would be simpler but less efficient. The way the code seems to work is calling `Add` X times before calling `Collate` Y times. So I think we could do without this `generateNewIndex` boolean.

However, to cover all cases where `Add` and `Collate` could be called in any orders, I decided to create this boolean to track when it is necessary to regenerate the index, because a version got added to the `version.VersionSlice`.

vervet-underground/internal/storage/revision.go
jlourenc: Sure

vervet-underground/internal/storage/revision.go
jlourenc: I htink `vervet-underground` could do with a larger refactor that would simplify the internal logic of each individual components. That'd be a bigger change that can be done at a later time.

ajbeattie-snyk: Cool - makes sense to me, thanks!

PR #314, opened 2023-11-03
chore(vu): bump go version in dockerfile

jlourenc:


vervet-underground/.golangci.yml
```diff
@@ -46,7 +46,6 @@ linters-settings:
   misspell:
     locale: US
   nolintlint:
-    allow-leading-space: true
```
jlourenc, line 49: No longer needed

vervet-underground/.golangci.yml
```diff
@@ -57,10 +56,8 @@ linters-settings:
   # ST1001: Dot imports that aren’t in external test packages are discouraged.
   stylecheck:
     checks: [ "all", "-ST1003" ]
-    go: "1.20"
```
jlourenc, line 60: No longer needed as it is derived from go.mod.

vervet-underground/.golangci.yml
```diff
@@ -57,10 +56,8 @@ linters-settings:
...
   staticcheck:
     checks: ["all"]
-    go: "1.20"
```
jlourenc, line 63: No longer needed as it is derived from go.mod.

VinceDeslo (approved):

PR #326, opened 2024-01-22
fix: prevent similar components from being overwritten

jgresty:
When collating schemas from multiple services, it is very likely that components with the same name will be defined in several places. We use a comparison function to detect when this is the case and inline components if they have already been declared.

This comparison function however ignored fields meant for documenting fields. Which is fine when it was written as the only concern was generating openapi specs inner service to validate requests. However now we are using the collated specs to generate documentation, this behaviour is very surprising. Especially when the description is context sensitive.

This will increase the size of our specs slightly, it should not be that significant. It will require us to recompile our existing specs to fix the broken versions.

### DIFF
--- a/collator.go
+++ b/collator.go
@@ -244,11 +244,7 @@ var cmpComponents = cmp.Options{
 	// Refs themselves can mutate during relocation, so they are excluded from
 	// content comparison.
 	cmp.FilterPath(func(p cmp.Path) bool {
-		switch p.Last().String() {
-		case ".Ref", ".Description", ".Example", ".Summary":
-			return true
-		}
-		return false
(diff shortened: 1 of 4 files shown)


jgresty:
@cmars  I didn't see any obvious reason why this simple patch would be an issue, however you did write the original code (albeit 2 years ago) so please shout if you can think of a distinct reason why we don't conflict on description/summary/example

tinygrasshopper:
it would be worth adding a unit test for this behaviour

jgresty:
> it would be worth adding a unit test for this behaviour

Updated the existing test to explicitly exercise this behaviour

tinygrasshopper (approved):

PR #336, opened 2024-03-04
fix: use correct docker image

corneliu-petrescu:
We're in a brownout, might as well do the upgrade to the "default" docker image in circle ci for build


jgresty (commented):
Unnecessary, default is used if no version is specified

PR #342, opened 2024-04-10
fix: throw error when conflicting components are found

jgresty:
When we roll up versions and copy over paths from older spec, we naïvely merge the components defined in those specs. This is a problem when similar named components are defined in two versions of a spec that is rolled up.

To fix this problem is a lot of effort which we do not have time for right now, but detecting it is relatively easy. This patch will fail building specs if we detect such a case so a user can work around it manually.

### DIFF
--- a/collator.go
+++ b/collator.go
@@ -4,6 +4,7 @@ import (
 	"fmt"
 	"regexp"
 	"sort"
+	"unicode"
 
 	"github.com/getkin/kin-openapi/openapi3"
 	"github.com/google/go-cmp/cmp"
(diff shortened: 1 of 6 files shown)


collator.go
```diff
@@ -244,7 +244,7 @@ var cmpComponents = cmp.Options{
 	// Refs themselves can mutate during relocation, so they are excluded from
 	// content comparison.
 	cmp.FilterPath(func(p cmp.Path) bool {
-		return p.Last().String() == ".Ref"
+		return p.Last().String() == ".Ref" || p.Last().String() == ".extra"
```
tinygrasshopper, line 247: should `.Ref`'s be ignored for deep compare? If objects refer to different types, the compare should fail in this context

merge.go
```diff
@@ -98,23 +104,50 @@ func mergeComponents(dst, src *openapi3.T, replace bool) {
...
+	return mergeMap(dst.Components.Callbacks, src.Components.Callbacks, replace)
 }
 
-func mergeMap[T any](dst, src map[string]T, replace bool) {
+func mergeMap[T any](dst, src map[string]T, replace bool) error {
 	for k, v := range src {
-		if _, ok := dst[k]; !ok || replace {
-			dst[k] = v
+		existing, exists := dst[k]
+		if exists && !replace && !componentsEqual(v, existing) {
```
tinygrasshopper, line 145: similar to the above comment; should we be using the same compare as the collator in this context? Should there be a specific comparator for this case, which compares ignoring the extra and fails for other conditions including if .Refs are the same

jgresty: This is the same compare.

tinygrasshopper (approved):

PR #347, opened 2024-06-12
Feat: merge vervet unground into main modules

jgresty:
Having vervet underground in a separate package means that it is a chore to update the version of vervet it uses, as we need to release vervet and then update the references in vervet underground after. This makes working on this repo harder than it needs to be.

This merges the packages so every release of vervet will also release vervet underground.

There shouldn't be anything too surprising in this diff, it is mostly moving files and updating import paths. The entry points have changed from api and scraper to vu-api and vu-scraper for clarity.

Direct dependencies have had their versions preserved, however keeping all the indirect dependency versions consistent would be far too much effort so this resolves them again - from `go mod tidy`.

READMEs have been concatenated, that might need to be revisited...

Note that this bumps the package version to v7 as there was previously a breaking change released which bumped our git tag, however we didn't align the go package version so it was in a broken state. As this is also a breaking change I elected to roll everything together here to a final v7 version, there are no more expected breaking changes for v7 so we can use it as the new recommended version.


RaduPetreTarean (approved):
Brilliant work!

PR #351, opened 2024-07-11
feat: do not collate sunset paths

RaduPetreTarean:
This PR introduces a feature to filter out API paths marked as sunset eligible in the Vervet tool's collation process. It ensures that paths with the x-snyk-sunset-eligible attribute, dated earlier than the current date, are excluded from the final collated specifications.

### DIFF
--- a/internal/storage/collator.go
+++ b/internal/storage/collator.go
@@ -4,6 +4,7 @@ import (
 	"fmt"
 	"net/url"
 	"sort"
+	"time"
 
 	"github.com/getkin/kin-openapi/openapi3"
 	"github.com/rs/zerolog/log"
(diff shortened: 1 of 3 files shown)


internal/storage/collator.go
```diff
@@ -126,6 +144,11 @@ func (c *Collator) Collate() (map[vervet.Version]openapi3.T, error) {
 				collatorMergeError.WithLabelValues(version.String()).Inc()
 				return nil, err
 			}
+			for path, pathItem := range spec.Paths {
+				if isPathSunsetEligible(pathItem) {
+					delete(spec.Paths, path)
```
tinygrasshopper, line 149: I think this change does need a feature flag/pivot date check

PR #352, opened 2024-07-17
feat: skip experimental versions on scrape

RaduPetreTarean:
This PR changes the functionality of the scraper by making it skip experimental versions resulting in docs no longer generating for them.


jgresty:
I feel like there are some documentation changes that are also needed, not sure where they would be though

PR #353, opened 2024-07-22
feat: add new codepath for building specs 

jgresty:
This codepath is an attempt to simplify the existing build process. It
is a clean room implementation to avoid adopting any unnecessary steps
from the old flow. Implemented as a new command to make it work
alongside the old command.

So far it only merges paths and not any other part of the spec. This
will need to be worked on before it is suitable, however we can begin
testing specs with this new command by releasing in progress versions.

Eventually this will supplant the old flow when we are confident that it
is feature complete.


.golangci.yml
```diff
@@ -33,7 +33,6 @@ linters:
     - errname
     # - errorlint - Good to have linter but potential to introduce breaking changes.
     # - exhaustive - Causes too much noise at the moment.
-    - exportloopref
```
tinygrasshopper, line 36: what the removal of the lint, intentional?

tinygrasshopper (approved):

tinygrasshopper:
Cursory read seems good; happy to merge it in and test/built on top of.

Thoughts about the linter?

.golangci.yml
jgresty: Yes see commit message, this is no longer reinvent in go 1.22: https://go.dev/blog/loopvar-preview

.golangci.yml
```diff
@@ -43,7 +42,6 @@ linters:
     - gocyclo
     # - gosec - Good to have linter but potential to introduce breaking changes.
     # - forbidigo - Good to have linter but potential to introduce breaking changes.
-    - ifshort
```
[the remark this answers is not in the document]
jgresty: Yes this linter has been deactivated in this version of golangci-lint and will error the process if enabled

PR #355, opened 2024-08-06
feat: breaking changes check

RaduPetreTarean:
This PR adds a check to the simple build flow to ensure that all sequential versions introduce breaking changes. 

If a version does not introduce a breaking change then that version should be "rolled into" the previous version as it is non breaking and the changes can be added in without breaking existing flows.

### DIFF
--- a/internal/simplebuild/build.go
+++ b/internal/simplebuild/build.go
@@ -2,7 +2,10 @@ package simplebuild
 
 import (
 	"context"
+	"errors"
 	"fmt"
+	"os"
+	"os/exec"
 	"path/filepath"
 	"slices"
 	"time"
@@ -16,6 +19,9 @@ import (
 
 func Build(ctx context.Context, project *config.Project) error {
 	for _, apiConfig := range project.APIs {
+
+		fmt.Printf("Processing API: %s\n", apiConfig.Name)
+
 		operations, err := LoadPaths(ctx, apiConfig)
 		if err != nil {
 			return err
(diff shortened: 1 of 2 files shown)


internal/simplebuild/build.go
```diff
@@ -234,3 +259,47 @@ func (vs VersionSet) Annotate() {
...
+
+		// Write specs to temporary files
+		if err := writeTempSpecFile(prevDoc.Doc, prevSpecPath); err != nil {
+			return err
+		}
+		if err := writeTempSpecFile(currDoc.Doc, currSpecPath); err != nil {
+			return err
+		}
+
+		cmd := exec.Command("oasdiff", "breaking", "--fail-on", "ERR", prevSpecPath, currSpecPath)
```
tinygrasshopper, line 280: this would need to be a code dependency, otherwise we would have to get everyone to install `oasdiff`

internal/simplebuild/build.go
```diff
@@ -234,3 +258,47 @@ func (vs VersionSet) Annotate() {
...
 }
+
+func CheckBreakingChanges(docs DocSet) error {
+	for i := 1; i < len(docs); i++ {
+		prevDoc := docs[i-1]
+		currDoc := docs[i]
+
+		// Create temporary file paths for previous and current specs
+		prevSpecPath := filepath.Join(os.TempDir(), fmt.Sprintf("spec-%d.json", i-1))
+		currSpecPath := filepath.Join(os.TempDir(), fmt.Sprintf("spec-%d.json", i))
```
tinygrasshopper, line 269: These specs are written to files later down, we could validate them there.

Also I think oasdiff can take strings as well, when using it as a golang lib

tinygrasshopper:
We should add a unit test for simplebuild to validate this one

jgresty:
Does #365 make this redundant?

RaduPetreTarean:
Yes, closing this PR

PR #357, opened 2024-08-13
fix: upgrade to latest kin-openapi [try-2]

tinygrasshopper:


remove_elements.go
```diff
@@ -65,45 +63,31 @@ func RemoveElements(doc *openapi3.T, excludes ExcludePatterns) error {
...
-// StructField implements reflectwalk.StructWalker.
-func (ex *excluder) StructField(field reflect.StructField, v reflect.Value) error {
-	if field.Name != "Extensions" || !v.CanInterface() {
-		return nil
-	}
-
-	switch v.Interface().(type) {
-	case map[string]interface{}:
-		ex.applyExtensions(v.Addr().Interface().(*map[string]interface{}))
+	for _, pathItem := range ex.doc.Paths.Map() {
```
jgresty, line 66: This is a very narrow use case but this only covers a small part of the doc compared to the old implementation

resource.go
```diff
@@ -331,14 +337,37 @@ func loadResource(specPath string, versionStr string) (*ResourceVersion, error)
...
-	doc.InternalizeRefs(ctx, nil)
+	doc.InternalizeRefs(ctx, ResolveRefsWithoutSourceName)
 	return doc.ResolveRefs()
 }
+
+// ResolveRefsWithoutSourceName resolves references without the source url/file name in ref
+// background: this was the way kin-openapi used to resolve references, but it was changed
+// in the recent versions(v0.127.0) to include the filename in the ref name. Although this
+// method prevents conflicts, it causes existing specs to break.
+func ResolveRefsWithoutSourceName(t *openapi3.T, componentRef openapi3.ComponentRef) string {
```
jgresty, line 359: I can't clearly see what this is doing which implies some tests would be appreciated to use as examples

jgresty (approved):
This is a lot clearer than it was before however I don't really like the amount of repetitive code. I think there could be a middle ground with some light reflection since all of these types we are operating over have the same `.Ref` and `.Value` members which is all we care about.

Not going to block on that but it is an area we could improve on if we find this is a source of future bugs

remove_elements.go
tinygrasshopper: agreed; I would argue the old implementation going much wider than intended. I was testing with the existing specs didnt find a use for replacing other extensions

PR #364, opened 2024-08-16
Fix/conflict components merge main

tinygrasshopper:


internal/simplebuild/refs.go
```diff
@@ -55,59 +55,138 @@ func NewRefResolver() refResolver {
...
-// documents make sure that any other documents are serialised before resolving
-// refs. This method only ensures the current document is correct.
-func (rr *refResolver) ResolveRefs(doc *openapi3.T) error {
-	// Refs use a full path eg #/components/schemas/..., to avoid having a
-	// special case at the top level we pass the entire document and trust the
-	// refs to not reference parts of the document they shouldn't.
-	rr.doc = doc
-	return reflectwalk.Walk(doc.Paths, rr)
+func (rr *refResolver) deRef(orignalRef string, component any) (string, error) {
+	newRef, err := rr.deref(orignalRef, reflect.ValueOf(component))
```
jgresty, line 59: We now have method called `deRef` and `deref`, perhaps something like `copyToComponents` would be better?

tinygrasshopper: ack

jgresty (approved):
Thanks for this change, besides the method names it looks good to me

PR #365, opened 2024-08-16
Feat/breaking changes check with oasdiff lib

tinygrasshopper:


corneliu-petrescu:
It would be great to have an integration test

corneliu-petrescu (approved):

PR #368, opened 2024-08-22
fix: update sunset header

corneliu-petrescu:
Updates sunset header with the correct description as per implementation

### DIFF
--- a/internal/storage/collator.go
+++ b/internal/storage/collator.go
@@ -131,6 +131,10 @@ func (c *Collator) Collate() (map[vervet.Version]openapi3.T, error) {
 				collatorMergeError.WithLabelValues(version.String()).Inc()
 				return nil, err
 			}
+
+			// Overrides sunset header documentation until we can provide a more definitive fix
+			overrideSunsetHeader(spec)
+
 			if err := c.applyOverlay(spec); err != nil {
 				log.Error().Err(err).Msgf("failed to merge overlay for version %s", version)
 				collatorMergeError.WithLabelValues(version.String()).Inc()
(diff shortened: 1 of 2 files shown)


internal/storage/collator.go
```diff
@@ -143,6 +147,44 @@ func (c *Collator) Collate() (map[vervet.Version]openapi3.T, error) {
...
+				if responses.Value.Headers["sunset"] != nil {
+					if responses.Value.Headers["sunset"].Value == nil {
+						continue
+					}
+					responses.Value.Headers["sunset"].Value.Description = headerDescription
+
+					if responses.Value.Headers["sunset"].Value.Schema.Value == nil {
+						continue
+					}
+					responses.Value.Headers["sunset"].Value.Example = example
```
jgresty, line 173: This should go above the schema value check

internal/storage/collator.go
```diff
@@ -143,6 +147,44 @@ func (c *Collator) Collate() (map[vervet.Version]openapi3.T, error) {
...
+					responses.Value.Headers["sunset"].Value.Schema.Value.Format = schemaFormat
+				}
+			}
+		}
+	}
+
+	if doc.Components != nil && doc.Components.Headers["sunsetHeader"] != nil {
+		if doc.Components.Headers["sunsetHeader"].Value == nil {
```
jgresty, line 181: Should be `SunsetHeader` (upper case S) both cases here

jgresty (changes requested):
You're doing `== nil` as lot to check if a map key exists then reading it again after. The code you have here will only work if the values in the map are pointers, and I'm not going to check all cases here. The typical way to do this In go is `if val, ok := myMap["foo"]; ok {`

https://stackoverflow.com/questions/2050391/how-to-check-if-a-map-contains-a-key-in-go

jgresty (approved):

PR #369, opened 2024-08-26
feat: single-version-resources to be before latest version

RaduPetreTarean:
This PR attempts to add a check during simple build to ensure that resources which have only one version (new resources since simplebuild is only run after the pivot date) are versioned using the latest release of the global API ensuring they do not end up creating a new version of the global API

### DIFF
--- a/internal/simplebuild/build.go
+++ b/internal/simplebuild/build.go
@@ -399,8 +399,14 @@ func CheckSingleVersionResourceToBeBeforeLatestVersion(paths []string, latestVer
 			}
 
 			if version.Date.After(latestVersion.Date) {
-				return fmt.Errorf("version %s is after the last released version of the global API %s. Please change the version date to be before %s or at the same date",
-					version.Date.Format("2006-01-02"), latestVersion.Date.Format("2006-01-02"), latestVersion.Date.Format("2006-01-02"))
+				return fmt.Errorf(
+					"version %s is after the last released version of the global API %s. "+
+						"Please change the version date to be before %s or at the same date",
+					version.Date.Format("2006-01-02"),
+					latestVersion.Date.Format("2006-01-02"),
+					latestVersion.Date.Format("2006-01-02"),
+				)
+
 			}
 		}
 	}
--- a/internal/simplebuild/build.go
+++ b/internal/simplebuild/build.go
@@ -2,9 +2,14 @@ package simplebuild
 
 import (
 	"context"
+	"encoding/json"
 	"fmt"
+	"io"
+	"net/http"
 	"path/filepath"
 	"slices"
+	"sort"
+	"strings"
 	"time"
 
 	"github.com/getkin/kin-openapi/openapi3"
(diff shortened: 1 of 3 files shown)


jgresty (commented):
This will prevent any new versions from being created, even if there is a breaking change.

Also this repo is open source so may be used outside of Snyk, this patch will mean by default anyone running this tool will be tied to our versioning dates.

internal/simplebuild/build.go
```diff
@@ -314,3 +342,67 @@ func CheckBreakingChanges(docs DocSet) error {
...
+
+	var dates = make([]string, 0, len(versions))
+
+	for _, version := range versions {
+		parts := strings.Split(version, "~")
+		dates = append(dates, parts[0])
+	}
+	sort.Strings(dates)
+
+	latestVersion, err := vervet.ParseVersion(dates[len(dates)-1])
```
jgresty, line 376: Since we just want the latest, a linear search will be much faster than sorting the whole list

internal/simplebuild/build.go
```diff
@@ -314,3 +342,67 @@ func CheckBreakingChanges(docs DocSet) error {
...
+		if len(versions) == 1 {
+			versionStr := versions[0]
+			version, err := vervet.ParseVersion(versionStr)
+			if err != nil {
+				return fmt.Errorf("invalid version %q", versionStr)
+			}
+
+			if version.Date.After(latestVersion.Date) {
+				return fmt.Errorf("version %s is after the last released version %s",
+					version.Date.Format("2006-01-02"), latestVersion.Date.Format("2006-01-02"))
```
jgresty, line 403: Would be useful to tell the user what to do, eg set the resource version to a specific date

internal/simplebuild/build.go
```diff
@@ -314,3 +342,67 @@ func CheckBreakingChanges(docs DocSet) error {
...
+
+	latestVersion, err := vervet.ParseVersion(dates[len(dates)-1])
+	if err != nil {
+		return vervet.Version{}, fmt.Errorf("failed to parse latest version date %q: %w", dates[len(dates)-1], err)
+	}
+
+	return latestVersion, nil
+}
+
+func CheckSingleVersionResource(paths []string, latestVersion vervet.Version) error {
```
jgresty, line 384: It took me a while to understand this patch and it is mostly down to this name, which could be more descriptive. What are we checking for?

internal/simplebuild/build.go
```diff
@@ -19,16 +24,39 @@ import (
...
+	project *config.Project,
+	startDate vervet.Version,
+	versioningUrl string,
+	appendOutputFiles bool,
+) error {
 	if time.Now().Before(startDate.Date) {
 		return nil
 	}
+
+	latestVersion, err := fetchLatestVersion(versioningUrl)
```
jgresty, line 38: This is using the same API versions across all of the defined API types and fetching that from a single API (the public one). Was this a deliberate decision to tie the other API types (eg hidden and internal) to the public API versions?

jgresty (approved):

internal/simplebuild/build.go
```diff
@@ -314,3 +342,72 @@ func CheckBreakingChanges(docs DocSet) error {
...
+		return vervet.Version{}, fmt.Errorf("failed to parse versioning information: %w", err)
+	}
+
+	var dates = make([]string, 0, len(versions))
+
+	for _, version := range versions {
+		parts := strings.Split(version, "~")
+		dates = append(dates, parts[0])
+	}
+	sort.Strings(dates)
```
tinygrasshopper, line 374: dates are pre-sorted in the response

PR #376, opened 2024-09-04
Feat/sunset policy

RaduPetreTarean:
This PR attempts to add a sunset policy following to be used for filtering out sunset versions.


jgresty (commented):
This is already defined in `version.go`

PR #379, opened 2024-09-23
feat: annotate operations with stability level

RaduPetreTarean:
This PR attempts to annotate the operations added during the new build process with the x-stability-level and x-snyk-api-stability annotations. These annotations are meant for the client to be able to tell which stability the requested path is at.

### DIFF
--- a/internal/simplebuild/build.go
+++ b/internal/simplebuild/build.go
@@ -294,12 +294,18 @@ func (vs VersionSet) Annotate() {
 
 	for idx, op := range vs {
 		if op.Operation.Extensions == nil {
-			op.Operation.Extensions = make(map[string]interface{}, 6)
+			op.Operation.Extensions = make(map[string]interface{}, 8)
 		}
 		op.Operation.Extensions[vervet.ExtSnykApiResource] = op.ResourceName
 		op.Operation.Extensions[vervet.ExtSnykApiVersion] = op.Version.String()
 		op.Operation.Extensions[vervet.ExtSnykApiReleases] = releases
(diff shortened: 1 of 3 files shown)


internal/simplebuild/build.go
```diff
@@ -294,12 +294,18 @@ func (vs VersionSet) Annotate() {
...
 		}
 		op.Operation.Extensions[vervet.ExtSnykApiResource] = op.ResourceName
 		op.Operation.Extensions[vervet.ExtSnykApiVersion] = op.Version.String()
 		op.Operation.Extensions[vervet.ExtSnykApiReleases] = releases
 		op.Operation.Extensions[vervet.ExtSnykApiLifecycle] = op.Version.LifecycleAt(time.Time{}).String()
+
+		if op.Version.Stability == vervet.StabilityBeta {
+			op.Operation.Extensions[vervet.ExtApiStabilityLevel] = "beta"
+			op.Operation.Extensions[vervet.ExtSnykApiStability] = "beta"
+		}
```
jgresty, line 307: Is there any reason why this is only done for beta?
```suggestion
		op.Operation.Extensions[vervet.ExtApiStabilityLevel] = op.Version.Stability.String()
		op.Operation.Extensions[vervet.ExtSnykApiStability] = op.Version.Stability.String()
```

RaduPetreTarean: I just pushed a commit to do it for all stabilities. initially went with beta as it is the only stability other than GA at the moment but it makes sense to add it for both.

tinygrasshopper (approved):

PR #381, opened 2024-09-30
fix: map x-stability value to a parsable value

tinygrasshopper:
context: 
the x-stability-level values dont map directly to vervet stability levels; this PR maps "ga" to "stable"

If we dont fix this, tools like oas-diff start failing the generated spec


jgresty (approved):
Well that is annoying, any chance we could change the values upstream so the "stability" term it is consistent within all of Snyk?

tinygrasshopper:
i feel changing vervet and the historical values might be a bigger lift; I see diminishing returns in that refactor

PR #384, opened 2024-10-15
fix: add default values to openapi docs

jgresty:
In the event that there are endpoints defined for a version but nothing to collate, eg when all the endpoints are experimental, then we will output a stub document with null values for OpenAPI and Paths. This stub document cannot be parsed by cerberus, which will then crash on startup.

This adds default values so that cerberus can parse the document, even though there are no routable paths for this version.


tinygrasshopper:
is there a way to validate those are the only ones needed by cerberus?

jgresty:
> is there a way to validate those are the only ones needed by cerberus?

We now call `Validate` before writing, which is the same function cerberus was erroring out on when loading.

tinygrasshopper (approved):

PR #387, opened 2024-10-17
fix: prefer cerberus resolved version in versionware

tinygrasshopper:
- Update the middleware to choose the `snyk version served` header when found
- Header set in https://github.com/snyk/cerberus/blob/1e5cd96dbe5b807106b4c41b3a0983c1e61805fe/internal/filter/versionheader.go#L29


tinygrasshopper:
Version information from cerb is not percolated down today

PR #389, opened 2024-10-17
fix: update versionware to consider pivot date

tinygrasshopper:
- fixes resolution for beta endpoints, post pivot date
- beta endpoints are now folded into the GA spec
- not resolving `beta` and `experimental` versions post pivot date

### DIFF
--- a/version.go
+++ b/version.go
@@ -324,6 +324,26 @@ func (vi *VersionIndex) Resolve(query Version) (Version, error) {
 	return Version{}, ErrNoMatchingVersion
 }
 
+// ResolveGAorBetaStability returns the GA or beta version effective on the query version date at
+// the given version date. Returns ErrNoMatchingVersion if no version matches or if query
+// stability is not GA.
+func (vi *VersionIndex) ResolveGAorBetaStability(query Version) (Version, error) {
+	i, err := vi.resolveIndex(query.Date)
+	if query.Stability != StabilityGA {
+		return Version{}, ErrNoMatchingVersion
+	}
+	if err != nil {
+		return Version{}, err
+	}
+	if stabDate := vi.effectiveVersions[i].stabilities[StabilityGA]; !stabDate.IsZero() {
+		return Version{Date: stabDate, Stability: StabilityGA}, nil
+	}
+	if stabDate := vi.effectiveVersions[i].stabilities[StabilityBeta]; !stabDate.IsZero() {
+		return Version{Date: stabDate, Stability: StabilityBeta}, nil
+	}
+	return Version{}, ErrNoMatchingVersion
+}
+
 // Versions returns each Version defined.
 func (vi *VersionIndex) Versions() VersionSlice {
 	vs := make(VersionSlice, len(vi.versions))
@@ -498,3 +518,6 @@ func defaultLifecycleAt() time.Time {
 	}
 	return timeNow().UTC()
 }
+
+// DefaultPivotDate is the default pivot date after which the versioning strategy changes.
+var DefaultPivotDate = MustParseVersion("2024-10-15")
--- a/version_test.go
+++ b/version_test.go
@@ -2,6 +2,7 @@ package vervet_test
 
 import (
 	"fmt"
+	"reflect"
 	"testing"
 	"time"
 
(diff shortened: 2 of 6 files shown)


version_test.go
```diff
@@ -437,3 +438,98 @@ func TestLifecycleAtDefaultDate(t *testing.T) {
...
+			versions: VersionSlice{
+				MustParseVersion("2021-06-07"),
+				MustParseVersion("2021-06-08~beta"),
+				MustParseVersion("2021-06-08~beta"),
+				MustParseVersion("2021-06-09"),
+				MustParseVersion("2021-06-10~beta"),
+			},
+			query:   MustParseVersion("2021-06-12"),
+			want:    MustParseVersion("2021-06-09"),
+			wantErr: false,
```
jgresty, line 499: This is exactly how the old resolver worked. This patch does nothing to actually resolve the issue.

In this case we need to resolve to the `2021-06-10~beta` spec.

tinygrasshopper: those cases are tested below in the handler test, the handler is aware of the pivot date:
https://github.com/snyk/vervet/pull/389/files/c842e9a5372d65981db5ccc081d0cda092bcba0e#diff-b0d41e8c9c15adfca870b5fc5afb2a2afedd3d6e595bcc00988ad1dca1ac6627R125-R137

`ResolveGAStability` fails(https://github.com/snyk/vervet/pull/389/files/c842e9a5372d65981db5ccc081d0cda092bcba0e#diff-2e0e5f9157b8e88ebe27e98732cd70b20a2ab48d5946caaa6788a7742ce4c45fR508) if the passed in version is a non ga version

happy to add a `2021-06-10~beta` spec test case, but that wont make a difference, as it will fail

jgresty: Handlers are defined at the beta spec. Requests are made without a stability so are considered ga.

A developer makes an endpoint and the only version is `2024-12-01~beta`, a user makes a request using `version=2024-12-05`, this will not resolve that request to the beta handler so the request will fail with a not found error.

jgresty: Made this comment in the wrong place, I will annotate the right code...

version.go
```diff
@@ -324,6 +324,23 @@ func (vi *VersionIndex) Resolve(query Version) (Version, error) {
 	return Version{}, ErrNoMatchingVersion
 }
 
+// ResolveGAStability returns the GA version effective on the query version date at
+// the given version date. Returns ErrNoMatchingVersion if no version matches or if query
+// stability is not GA.
+func (vi *VersionIndex) ResolveGAStability(query Version) (Version, error) {
```
jgresty, line 330: Inbound requests after the pivot date do not have stability so will always be ga. Beta endpoint handlers will have the beta stability. We want if a request is made after the pivot date for it to resolve to the latest spec regardless of stability.

A developer makes an endpoint and the only version is `2024-12-01~beta`. A user makes a request with `version=2024-12-05`. That request should be served by the beta version as there are no other versions for that endpoint. This code will return NoMatchingVersion, exactly the same as the current resolver does.

tinygrasshopper: ah, gotcha; that is correct

tinygrasshopper: ok; changed the implementation to consider that, and added more handler tests to cover this scenario

version.go
```diff
@@ -324,6 +324,26 @@ func (vi *VersionIndex) Resolve(query Version) (Version, error) {
 	return Version{}, ErrNoMatchingVersion
 }
 
+// ResolveGAorBetaStability returns the GA or beta version effective on the query version date at
+// the given version date. Returns ErrNoMatchingVersion if no version matches or if query
+// stability is not GA.
+func (vi *VersionIndex) ResolveGAorBetaStability(query Version) (Version, error) {
+	i, err := vi.resolveIndex(query.Date)
+	if query.Stability != StabilityGA {
```
jgresty, line 332: So we are actively blocking users who still request `~beta`, this seems unnecessary and will annoy people trying to update their code quickly

version.go
```diff
@@ -324,6 +324,26 @@ func (vi *VersionIndex) Resolve(query Version) (Version, error) {
...
+	}
+	if err != nil {
+		return Version{}, err
+	}
+	if stabDate := vi.effectiveVersions[i].stabilities[StabilityGA]; !stabDate.IsZero() {
+		return Version{Date: stabDate, Stability: StabilityGA}, nil
+	}
+	if stabDate := vi.effectiveVersions[i].stabilities[StabilityBeta]; !stabDate.IsZero() {
+		return Version{Date: stabDate, Stability: StabilityBeta}, nil
+	}
```
jgresty, line 343: I think it would be better to just return the highest stability rather than specifically check for GA or Beta

versionware/handler.go
```diff
@@ -69,13 +69,24 @@ func (h *Handler) HandleErrors(errFunc VersionErrorHandler) {
...
+	} else {
+		resolvedVersion, err = h.index.ResolveGAorBetaStability(requested)
+		if err != nil {
+			return nil, nil, err
+		}
 	}
 	return &resolvedVersion, h.handlers[resolvedVersion], nil
 }
 
+var defaultPivotDate = vervet.MustParseVersion("2024-10-15")
```
jgresty, line 88: This should already be defined in vervet (internal/cmd/compiler.go). We should have one definition.

version.go
tinygrasshopper: this is intended, and as described in the new simplified versioning strategy: https://docs.google.com/document/d/1W8iGpAkefxfyZCzlEjbjn1GyfVByJBe21-a8mOcfabA/edit#heading=h.m2vuajiht7tt

I think its confusing to not have a spec with the `~beta` notation, and to pass it when calling the API. 

Only versioning on date is simpler. Also more practically, the beta versions are a small minority of our APIs right now.

versionware/handler.go
tinygrasshopper: this is fair, I was trying to avoid the pivot date going into the "core" vervet package, and only leave it at the edges, like the CLI and the handler. But will move it in for consistency

version.go
tinygrasshopper: This is explicitly stopping experimental use. 

Within beta and GA its effectively returning the highest stability version

version.go
RaduPetreTarean: Seems wrong to break peoples current implementations, i feel like we will get a lot of backlash from users if we block calls with ~beta

RaduPetreTarean: Maybe we should still allow it for a while and do a bigger announcement around breaking people using ~beta. I feel like users will say they were not given notice.

tinygrasshopper: The existing integrations will work as they do today, we dont be breaking users. 

When users update to a version `2024-10-15` they will see one specification `2024-10-15.json` not `2024-10-15~beta.yaml` like they had. So its seems symmetric to me to use `2024-10-15` as the version identifier as well.

RaduPetreTarean: Well if existing integrations are okay then its good. Sorry, just went off the comment.

version.go
jgresty: We could save ourselves some work on every request, and a bit of memory, by filtering them out when loading instead of doing it here

jgresty (approved):
Still don't think it is necessary to make this strict when we should already be preventing specs from being published upstream but I wont block on that

PR #391, opened 2024-10-18
fix: openapi walker walks into additional properties

tinygrasshopper:
- although we are enforcing additional properties `= false` going forward there are a few existing repos which use additional properties

### DIFF
--- a/internal/openapiwalker/walker.go
+++ b/internal/openapiwalker/walker.go
@@ -250,6 +250,14 @@ func ProcessRefs(data any, p RefProcessor) error {
 		if err := ProcessRefs(v.Not, p); err != nil {
 			return err
 		}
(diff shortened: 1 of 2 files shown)


jgresty (approved):
You sent me  down a rabbit hole of reading the spec. There are 3 different ways that `additionalProperties` is used:

 - If `properties` is defined on the same schema object then `additionalProperties` is a boolean allowing or disallowing properties that aren't listed on the spec
 - When `additionalProperties` is an inline value, the only key allowed is `type`, then the properties of the schema object are a map with unknown keys and the value type as defined by `additionalProperties`
 - Else when `additionalProperties` is a ref it represents exactly the same as `properties` but as a ref instead of inline
 
No idea why that last definition is needed when `properties` could have just been a ref type instead.

PR #396, opened 2024-11-08
Feat/two stage release

jgresty:
We are encouraging people to pull this tool using `go run ...` which fetches
the source and builds on the users system. This means that --version would
always return "develop". For this to have a useful version number we need to
bake the version number into the source.

We can do this by releasing slightly differently; instead of tagging the head
of main each time something is merged in we can push a new commit which
contains the version number then tag that commit. We then perform the rest of
the release (docker container, npm package) on that tagged commit instead of
the merged one.

This will allow us to assert that a source build is the same as a binary build
and allow accurate version information.


.circleci/config.yml
```diff
@@ -94,18 +97,36 @@ jobs:
...
-            ./scripts/dist.bash
-            ./scripts/release.bash
+            ./scripts/genversion.bash
+            go run github.com/nix-community/gomod2nix@latest
+            if [ -z $(git config user.email) ]; then
+                git config credential.helper 'cache --timeout=120'
+                git config user.email "vervet-ci@noreply.snyk.io"
+                git config user.name "Vervet CI"
+            fi
+            git add internal/cmd/cmd.go gomod2nix.toml
```
tinygrasshopper, line 115: I dont understand this transformation, can we add a comment to explain what `gomod2nix` is doing in this context

.circleci/config.yml
```diff
@@ -50,10 +57,6 @@ jobs:
   test:
     <<: *test_defaults
     steps:
-      - node/install
-      - run:
-          name: Install spectral
-          command: npm install -g @stoplight/spectral@6.5.0
```
tinygrasshopper, line 56: good catch

tinygrasshopper:
seems like a lot of orchestration to get the current version in. 

If we want the correct version to be displayed with `go run ...`, we can set the default value to be `svu.Current()`(which confusingly is the equivalent of `svu next` on the command lint) using  rather than `develop`, and that should get the result we intend

tinygrasshopper:
just to be concrete, this is what I mean:
```diff
diff --git a/internal/cmd/cmd.go b/internal/cmd/cmd.go
index 4f9a30b..b3dc1d7 100644
--- a/internal/cmd/cmd.go
+++ b/internal/cmd/cmd.go
@@ -5,11 +5,13 @@ import (
 	"context"
 	"errors"
 	"fmt"
+
 	"io"
 	"os"
 	"path/filepath"
 	"strings"
 
+	"github.com/caarlos0/svu/v2/pkg/svu"
 	"github.com/manifoldco/promptui"
 	"github.com/urfave/cli/v2"
 )
@@ -65,7 +67,7 @@ func NewApp(app *cli.App, vp VervetParams) *VervetApp {
 var CLIApp = cli.App{
 	Name:    "vervet",
 	Usage:   "OpenAPI resource versioning tool",
-	Version: "develop", // Set in init created with go generate.
+	Version: currentVersion(),
 	Flags: []cli.Flag{
 		&cli.BoolFlag{
 			Name:  "debug",
@@ -85,6 +87,15 @@ var CLIApp = cli.App{
 	},
 }
 
+func currentVersion() string {
+	current, err := svu.Current()
+	if err != nil {
+		//cases in which git cant be found
+		return "develop"
+	}
+	return current
+}
+
 // Prompt is the default interactive prompt for vervet.
 type Prompt struct{}
```

jgresty:
> seems like a lot of orchestration to get the current version in.
> 
> If we want the correct version to be displayed with `go run ...`, we can set the default value to be `svu.Current()`(which confusingly is the equivalent of `svu next` on the command line) using rather than `develop`, and that should get the result we intend

`svu.Current()` relies on git tags being available, `go get` does not preserve any information from this git repository; it just downloads the tagged snapshot.

.circleci/config.yml
jgresty: It does one thing: 
> From the Go project directory execute:
> $ gomod2nix
> This will create gomod2nix.toml

https://github.com/nix-community/gomod2nix

RaduPetreTarean (approved):

PR #401, opened 2024-11-26
fix: add a routable string helper to verevet.Version

tinygrasshopper:
- helps with templating version string in the generators, for example here: https://github.com/snyk/rest-node-generate/blob/628f7d200ac16d1d97cb9c99c90e3444c41813ef/lib/rpc/routes.ts.tmpl#L28
- tests the logic in vervet rather than adding an `if` in the templating


version.go
```diff
@@ -34,6 +34,15 @@ func (v Version) String() string {
...
 }
 
+// RoutableString returns the string representation of the version that can used while setting up the router.
+// If the version is before the DefaultPivotDate, it will return the stability in the string.
+func (v Version) RoutableString() string {
+	if v.Date.Before(DefaultPivotDate.Date) {
+		return v.String()
+	}
+	return v.DateString()
+}
```
jgresty, line 44: Should this replace `String`? What use cases are there for using the original `String` method over this one?

tinygrasshopper: The blast radius for that change would be quite a lot, the `.String` method is used in other contexts. We can replace the `.String` -> `.RoutableString` iteratively where needed

jgresty: That's what semantic versioning is for

tinygrasshopper: > That's what semantic versioning is for

Dont get what you mean by that bit

Currently, there are 58 usages for `version.String()` in the vervet codebase and the `RoutableString` implementation is not valid for all of them. We might introduce a lot of unintended side-effects if we replace the String method

PR #403, opened 2025-01-22
feat: sanitize-command

RaduPetreTarean:
This PR introduces a new sanitize command to remove sensitive information from versioned OpenAPI specs before distribution. 

Given a directory of compiled specs (one subfolder per version), the command loads each spec.yaml, filters out configurable extensions, headers, or entire paths, and saves the sanitized spec to a specified output directory.

```
vervet sanitize \
  --compiled-path ./internal/rest/api/versions \
  --out ./sanitized \
  --exclude-extension '^x-internal$' \
  --exclude-header '^X-Secret$' \
  --exclude-path '/internal'
```

This will read all spec.yaml files under each version subfolder, remove fields matching the exclusion criteria, and write sanitized openapi.yaml files to the ./sanitized directory.

### DIFF
--- a/internal/cmd/sanitize.go
+++ b/internal/cmd/sanitize.go
@@ -1,0 +1,151 @@
+package cmd
+
+import (
+	"fmt"
+	"os"
+	"path/filepath"
+	"sort"
+
+	"github.com/urfave/cli/v2"
+
+	"github.com/snyk/vervet/v8"
+	"github.com/snyk/vervet/v8/internal/storage"
+)
+
+var SanitizeCommand = cli.Command{
+	Name:      "sanitize",
+	Usage:     "Manually load compiled specs from subfolders, strip sensitive fields, write sanitized output",
+	ArgsUsage: " ",
+	Flags: []cli.Flag{
+		&cli.StringFlag{
+			Name:     "compiled-path",
+			Required: true,
+			Usage:    "Directory containing subfolders for each version, e.g. internal/rest/api/versions/",
+		},
+		&cli.StringFlag{
+			Name:     "out",
+			Usage:    "Output directory for sanitized versions",
+			Value:    "sanitized",
+			Required: false,
+		},
+		&cli.StringSliceFlag{
+			Name:  "exclude-extension",
+			Usage: "Regex patterns of extension names to remove (repeatable)",
+		},
+		&cli.StringSliceFlag{
+			Name:  "exclude-header",
+			Usage: "Regex patterns of header names to remove (repeatable)",
+		},
+		&cli.StringSliceFlag{
+			Name:  "exclude-path",
+			Usage: "Exact path(s) to remove from the final OpenAPI specs",
+		},
+	},
+	Action: sanitizeAction,
+}
+
+func sanitizeAction(c *cli.Context) error {
+	excludePatterns := vervet.ExcludePatterns{
+		ExtensionPatterns: c.StringSlice("exclude-extension"),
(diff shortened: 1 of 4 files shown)


gablaxian:
```
--exclude-extension '^x-internal$' \
--exclude-header '^X-Secret$' \
--exclude-path '/internal'
```
Are these required? These seem unnecessary for this iteration if so, and make the command more complicated.

internal/cmd/sanitize.go
```diff
@@ -0,0 +1,151 @@
...
+		},
+		&cli.StringSliceFlag{
+			Name:  "exclude-path",
+			Usage: "Exact path(s) to remove from the final OpenAPI specs",
+		},
+	},
+	Action: sanitizeAction,
+}
+
+func sanitizeAction(c *cli.Context) error {
```
acaprar, line 47: we should add some tests for this

internal/cmd/sanitize.go
```diff
@@ -0,0 +1,151 @@
...
+			Name:     "compiled-path",
+			Required: true,
+			Usage:    "Directory containing subfolders for each version, e.g. internal/rest/api/versions/",
+		},
+		&cli.StringFlag{
+			Name:     "out",
+			Usage:    "Output directory for sanitized versions",
+			Value:    "sanitized",
+			Required: false,
+		},
```
acaprar, line 30: I think `compiled-path` and `out` will remain the same most of the time. How about we add a DEFAULT for each and only pass the flags when we want the `path` and `out` to be different.

This could also apply to the other flags as well.

internal/cmd/sanitize.go
```diff
@@ -0,0 +1,151 @@
...
+		return fmt.Errorf("failed to create out directory %q: %w", outDir, err)
+	}
+
+	versionDirs, err := findVersionDirs(compiledPath)
+	if err != nil {
+		return fmt.Errorf("failed to find version subfolders in %q: %w", compiledPath, err)
+	}
+
+	if len(versionDirs) == 0 {
+		fmt.Fprintf(c.App.Writer, "No version subfolders found in %q\n", compiledPath)
```
acaprar, line 67: looks like `Fprintf` has an error that isn't handled

RaduPetreTarean:
> ```
> --exclude-extension '^x-internal$' \
> --exclude-header '^X-Secret$' \
> --exclude-path '/internal'
> ```
> 
> Are these required? These seem unnecessary for this iteration if so, and make the command more complicated.

It allows the user to define which extensions/headers/paths etc he wants to strip out without us being explicit which ones. We could change this to be loaded from a separate config but the user will still have to define which ones as we cannot assume internal structure

internal/cmd/sanitize.go
RaduPetreTarean: While the out could be inferred from the compiled-path, i wanted to keep the utility away from the actual projects .vervet file such that it can be used by anyone on a simple directory of compiled specs by using the go run utility or similar

internal/cmd/sanitize.go
```diff
@@ -0,0 +1,151 @@
...
+			Required: true,
+			Usage:    "Directory containing subfolders for each version, e.g. internal/rest/api/versions/",
+		},
+		&cli.StringFlag{
+			Name:     "out",
+			Usage:    "Output directory for sanitized versions",
+			Value:    "sanitized",
+			Required: false,
+		},
+		&cli.StringSliceFlag{
```
tinygrasshopper, line 31: should add this config to the config file? in the sanitise block?

acaprar (approved):

PR #404, opened 2025-01-28
fix: continue build if fetching latest version fails

RaduPetreTarean:
This allows the build to pass if the server serving the latest global version is not available

### DIFF
--- a/internal/simplebuild/build.go
+++ b/internal/simplebuild/build.go
@@ -38,9 +38,10 @@ func Build(
 		return nil
 	}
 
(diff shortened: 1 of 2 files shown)


malinbranduse (commented):
looks good to me, only thing missing would be a unit test to confirm this edge case. might be helpful down the line to remember why this change was added and what's the expected behaviour.

malinbranduse (commented):
redacted

malinbranduse (approved):

PR #405, opened 2025-01-29
fix: truncate api names for old versions

RaduPetreTarean:
We previously appended the stability of apis before the pivot date, this is made after the truncation to 63 characters.

Which results in invalid names if api names are too long, this truncates the name back to 63 characters

### DIFF
--- a/internal/backstage/backstage.go
+++ b/internal/backstage/backstage.go
@@ -169,7 +169,7 @@ func LoadCatalogInfo(r io.Reader) (*CatalogInfo, error) {
 	if catalog.service != nil {
 		var apiNames []string
 		for _, apiName := range catalog.serviceComponent.Spec.ProvidesAPIs {
-			// Preserve manually added entries, things are are NOT vervet APIs
+			// Preserve manually added entries, things that are NOT vervet APIs
 			if _, ok := vervetAPINames[apiName]; !ok {
 				apiNames = append(apiNames, apiName)
 			}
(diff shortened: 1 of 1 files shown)


internal/backstage/backstage.go
```diff
@@ -304,6 +304,8 @@ func (c *CatalogInfo) vervetAPI(doc *vervet.Document, root string, pivotDate tim
 		}
 
 		name = name + "_" + version.Stability.String()
+		// Backstage names can only be a maximum of 63 characters
+		name = name[:63]
```
jgresty, line 308: This will cut off the version string, which will lead to conflicts. It should remove characters from the part of the name derived from  `doc.Info.Title`

internal/backstage/backstage.go
```diff
@@ -303,7 +303,13 @@ func (c *CatalogInfo) vervetAPI(doc *vervet.Document, root string, pivotDate tim
...
 		}
 
-		name = name + "_" + version.Stability.String()
+		stabilitySuffix := "_" + version.Stability.String()
+		availableTitleLen := 63 - len(name_suffix) - len(stabilitySuffix)
+		if availableTitleLen < 0 {
+			availableTitleLen = 0
+		}
+		name = fmt.Sprintf("%."+strconv.Itoa(availableTitleLen)+"s%s%s",
+			toBackstageName(doc.Info.Title), name_suffix, stabilitySuffix)
```
jgresty, line 312: If we are building the name from scratch again here then doing it above is redundant. You can remove the logic on line ~279.

Please also make sure you retain the comment that explains why we are doing this.

jgresty (approved):

